PR #13, opened 2022-03-23
set up thebe

OriolAbril:


aloctavodia:
When trying to run the interactive example i get `ModuleNotFoundError: No module named 'sklearn'`

OriolAbril:
This is a problem with preliz that this PR will fix, installing preliz doesn't install the dependencies because when installing preliz the requirements are taken from pyproject, not from requirements.txt. I have modified that so that installing preliz should now install the dependencies, now for testing for example, only requirements-dev and `.` need to be installed. 

I still haven't managed tkinter to work from binder, not sure it's possible to generate pop-ups from within binder :/, but the constraints function does work. I think ppa will work, we should add an example of each function to the notebook and I'll add pymc to the binder image, it looks like ppa needs a pymc model.

aloctavodia:
Cool, yes ppa needs a pymc model. And I am thinking on moving ppa also to tkinter as response should be faster, in particular if I we add some buttons.

PR #54, opened 2022-07-08
Miscelaneous changes

aloctavodia:
* Make `fit_moments` private to prevent possible confusions. The rationale for this method is  to provide a reasonable initial guess for optimization methods. So it is just an intermediate step and not intended to provide an accurate moment matching method.
* Improve description in the code of the initial guess heuristic.
* Improve warning message for maxent and quartile method
* Add example for the quartile method
* Fix typos in the notebook example.
* Other minor fixes.


preliz/distributions/continuous.py
```diff
@@ -126,7 +126,7 @@ def _update(self, lam):
         self.params = (self.lam,)
         self._update_rv_frozen()
 
-    def fit_moments(self, mean, sigma=None):  # pylint: disable=unused-argument
+    def _fit_moments(self, mean, sigma=None):  # pylint: disable=unused-argument
```
canyon289, line 129: If sigma is not used why not do this to make things very obvious?
```suggestion
    def _fit_moments(self, mean, _=None):  # pylint: disable=unused-argument
```

aloctavodia: In other parts of the code like in maxent function we call this generally as distribution._fit_moments(self, mean, sigma). So this avoids the need to handle special cases in those other parts.

PR #60, opened 2022-07-24
Roulette: Use ipywidgets instead of tkinter

### LINKED ISSUE #32: Tkinter not working with Thebe/Binder
Maybe we can make it work by doing 

```python
from xvfbwrapper import Xvfb

with Xvfb() as xvfb:
    pz.roulette()
```

or 

```python
import pyvirtualdisplay

_display = pyvirtualdisplay.Display(visible=False,  # use False with Xvfb
                                    size=(1400, 900))
_ = _display.start()
```

Both approaches require to have `xvfb` and maybe also `x11-utils`, which are not installed by default. I try to install them by adding a apt.txt file to the root of the repo. But it seems binder is not installing it.

### LINKED ISSUE #37: Explore ipywidgets and other jupyter/html5 friendly GUIs
We may want to keep Tkinter interface and offer more than one option.

aloctavodia:
closes #32
closes  #37


aloctavodia:
@OriolAbril any idea how to fix these versions incompatibilities?

OriolAbril:
Pip is messing up big time. RTD first installs `requirements-docs`, then installs PreliZ. 

On the requirements docs step:

```
Collecting nbclient<0.6,>=0.2
  Downloading nbclient-0.5.13-py3-none-any.whl (70 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 70.6/70.6 kB 214.0 MB/s eta 0:00:00
```

But then, when installing PreliZ, now nbclient is also an indirect dependency via ipywidgets, and somehow this happens:

```
Requirement already satisfied: nbclient>=0.5.0 in /home/docs/checkouts/readthedocs.org/user_builds/preliz/envs/60/lib/python3.9/site-packages (from nbconvert>=5->notebook>=4.4.1->widgetsnbextension~=3.6.0->ipywidgets->preliz==0.0.1.dev0) (0.5.13)
Collecting nbclient>=0.5.0
  Downloading nbclient-0.6.6-py3-none-any.whl (71 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 71.8/71.8 kB 211.0 MB/s eta 0:00:00
```

And at the end of the PreliZ install it even prints this:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
jupyter-cache 0.5.0 requires nbclient<0.6,>=0.2, but you have nbclient 0.6.6 which is incompatible.
```

I think adding `nbclient` as an explicit dependency for PreliZ will fix this. But I think the preferred solution should be making ipywidgets and scikit-learn optional dependencies, then install doing the following:

1. Install `requirements-optional`
2. Install `requirements-docs`
3. Install PreliZ

Extra requirements can be added in the [readthedocs.yml](https://github.com/arviz-devs/preliz/blob/main/.readthedocs.yaml#L11). It would look like

```
install:
   - requirements: requirements-optional.txt
   - requirements: requirements-docs.txt
```

aloctavodia:
BTW, we should get rid of Scikit-learn dependency is too big for what we need. So I will work in the next few weeks in making the ppa function usable and as part of that removing Scikit-learn.  But maybe we should try to use ipywidgets more and hence keep it as non-optional requirement... not sure.

PR #85, opened 2022-10-28
add Inverse Gamma

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -662,6 +662,77 @@ def rvs(self, size=1, random_state=None):  # pylint: disable=arguments-differ
...
+    .. plot::
+        :context: close-figs
+
+        import arviz as az
+        from preliz import InverseGamma
+        az.style.use('arviz-white')
+        alphas = [1., 2., 3.]
+        betas = [1., 1., .5]
+        for alpha, beta in zip(alphas, betas):
+            InverseGamma(alpha, beta).plot_pdf()
```
aloctavodia, line 686: This plot looks weird compared with the one from PyMC. Try with

```suggestion
            InverseGamma(alpha, beta).plot_pdf(support=(0, 3))
```

PR #97, opened 2022-11-13
Add alternative parametrization for normal distribution

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -974,25 +973,57 @@ class Normal(Continuous):
...
+
+        else:
+            raise ValueError(
+                "Incompatible parametrization. Either use mu " "and sigma, or mu and tau."
+            )
+
+        return mu, sigma, names
+
+    def _from_tau(self, tau):
+        sigma = 1 / tau ** (1 / 2)
```
aloctavodia, line 1021:
```suggestion
        sigma = 1 / tau **  0.5
```

preliz/distributions/continuous.py
```diff
@@ -1003,7 +1034,15 @@ def _get_frozen(self):
...
-        self.params = (self.mu, self.sigma)
+        self.tau = self._to_tau(sigma)
+
+        if self.param_names[1] == "sigma":
+            self.params_report = (self.mu, self.sigma)
+            self.params = (self.mu, self.sigma)
+        elif self.param_names[1] == "tau":
+            self.param_report = (self.mu, self.tau)
+            self.params = (self.mu, self.tau)
+
```
aloctavodia, line 1045:
```suggestion
        if self.param_names[1] == "sigma":
            self.params_report = (self.mu, self.sigma)
        elif self.param_names[1] == "tau":
            self.param_report = (self.mu, self.tau)
        self.params = (self.mu, self.sigma)

```

preliz/tests/test_distributions.py
```diff
@@ -61,6 +61,8 @@ def test_moments(distribution, params):
     dist = distribution(*params)
     if "student" in dist.name:
         dist_ = distribution(nu=params[0])
+    elif "normal" == dist.name:
+        dist_ = distribution(mu=params[0], sigma=params[1])
```
aloctavodia, line 65:
```suggestion
```

preliz/distributions/continuous.py
```diff
@@ -974,25 +973,57 @@ class Normal(Continuous):
...
+        if tau is None:
+            names = ("mu", "sigma")
+
+        elif tau is not None:
+            sigma = self._from_tau(tau)
+            names = ("mu", "tau")
+
+        else:
+            raise ValueError(
+                "Incompatible parametrization. Either use mu " "and sigma, or mu and tau."
```
aloctavodia, line 1015:
```suggestion
                "Incompatible parametrization. Either use mu and sigma, or mu and tau."
```

aloctavodia (changes requested):
Internally we should use only one parametrization, in this case `self.params = (self.mu, self.sigma)` simply because that was already the case.

PR #99, opened 2022-11-16
Add alternative parametrization for Student distribution

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -1243,33 +1252,62 @@ class Student(Continuous):
...
+        return lam
 
     def _get_frozen(self):
         frozen = None
         if any(self.params):
             frozen = self.dist(self.nu, self.mu, self.sigma)
         return frozen
 
-    def _update(self, mu, sigma, nu=None):
+    def _update(self, mu, sigma, nu=3):
```
aloctavodia, line 1298: nu should not be fixed

PR #106, opened 2022-11-21
Add alternative parametrization for halfnormal distribution

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -638,29 +638,59 @@ class HalfNormal(Continuous):
...
+
+    def _parametrization(self, sigma, tau):
+        if tau is None:
+            names = ("sigma",)
+
+        elif tau is not None:
+            sigma = self._from_tau(tau)
+            names = ("tau",)
+
+        else:
```
aloctavodia, line 682: We will never get into this branch, because tau is either None or not None. I think we should explicitly test that one is none and the other doesn't, then the opposite and then the else

PR #113, opened 2022-11-25
Add from_tau and to_tau functions to continuous module

### LINKED ISSUE #112: Define _from_tau and _to_tau only once
A few of the alternative parametrization involve converting from tau to sigma and vice-versa. We could define these functions only once in the library.

aleicazatti:
To define both functions only once in the library, close #112.

### DIFF
--- a/preliz/distributions/continuous.py
+++ b/preliz/distributions/continuous.py
@@ -684,14 +684,6 @@ class HalfNormal(Continuous):
 
         return sigma, names
 
-    def _from_tau(self, tau):
-        sigma = 1 / tau**0.5
-        return sigma
-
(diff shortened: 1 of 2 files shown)


aloctavodia:
I think the functions should be inside the continuous module, instead of being part of the base class. Mostly because not all continuous distributions need those methods.

preliz/distributions/continuous.py
```diff
@@ -16,6 +16,26 @@
 eps = np.finfo(float).eps
 
 
+class TauSigma:
```
aloctavodia, line 19: No need to define a class, just define the two functions and call them from within each distribution you need to.

PR #115, opened 2022-11-28
Add alternative parametrization for HalfStudent distribution

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -26,6 +26,16 @@ def to_tau(sigma):
     return tau
 
 
+def from_lam(lam):
```
aloctavodia, line 29: These expresions are equivalent to from_tau, to_tau. We should use them instead

PR #118, opened 2022-12-02
Add Logistic distribution

aleicazatti:


aloctavodia (commented):
It seems you added Testing_Preliz.ipynb by mistake

PR #120, opened 2022-12-05
Add Gumbel distribution

aleicazatti:


preliz/distributions/continuous.py
```diff
@@ -564,6 +564,81 @@ def _fit_mle(self, sample, **kwargs):
...
+
+       f(x \mid \mu, \beta) = \frac{1}{\beta}e^{-(z + e^{-z})}
+
+    where
+
+    .. math::
+
+        z = \frac{x - \mu}{\beta}
+
+     .. plot::
```
aloctavodia, line 583: This seems to be shifted by one space

PR #139, opened 2022-12-15
add docstring examples to maxent and quartile, update notebook to show how to fix parameters

aloctavodia:


aloctavodia:
@OriolAbril could you help me with this? 

I would like to show the code and plots for the examples in the docstring of maxent and quartile, and only the plots for https://preliz.readthedocs.io/en/latest/api_reference.html#module-preliz.distributions.continuous and https://preliz.readthedocs.io/en/latest/api_reference.html#module-preliz.distributions.discrete

preliz/maxent.py
```diff
@@ -40,6 +40,34 @@ def maxent(
...
+    See Also
+    --------
+    quartile : Find the distribution with the specified quartiles.
+
+    Examples
+    --------
+    Calculate the maxent Gamma distribution with 90 % of the mass between 1 and 8:
+
+    .. plot::
+        :context: close-figs
```
OriolAbril, line 53:
```suggestion
        :context: close-figs
        :include-source: true
```

I think given the situation you describe, the best option is to keep the default to false and add this to the code examples you want the source to be shown.

aloctavodia:
Thank you @OriolAbril!

PR #151, opened 2022-12-20
Add logo

aloctavodia:


OriolAbril:
el favicon para arviz me aparece bastante mas grande a mi: 

imatge

alguna idea de porque podria ser?

Re logo on the navbar, it should ideally be like this: https://github.com/arviz-devs/arviz/blob/main/doc/source/conf.py#L170 (using paths relative to the static path)

docs/conf.py
```diff
@@ -120,12 +126,12 @@
             "icon": "fab fa-github-square",
         },
     ],
+    "logo": {
+        "image_light": "PreliZ.png",
+        "image_dark": "PreliZ_white.png",
```
OriolAbril, line 131:
```suggestion
        "image_light": "PreliZ_flat.png",
        "image_dark": "PreliZ_flat_white.png",
```

I think this will look much better given the theme

aloctavodia: Sorry, I saw this after doing the last commit

docs/conf.py
```diff
@@ -94,13 +94,19 @@
...
+html_static_path = ["_static", "logos"]
+
 
 # -- Options for HTML output -------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
 html_theme = "pydata_sphinx_theme"
+html_favicon = "favicon.ico"
```
OriolAbril, line 109:
```suggestion
html_favicon = "logos/favicon.ico"
```

I don't know why but the favicon is not relative to the static path 🤷🏿

OriolAbril (approved):

PR #178, opened 2023-01-10
Add nbsphinx

aloctavodia:


OriolAbril:
you should not use nbsphinx and myst-nb at the same time, only one can be actually used to render notebooks. Both should support rendering ipywidgets though. I'll take a look at why myst-nb is not working correctly

aloctavodia:
Thanks, the only way to render the widgets I have found is to run a notebook and then click on  "save widget state". But I think this has to be done every time the state changes, so I was trying to see in nbsphix was able to do that automatically. Do you know if myst-nb can do it automatically?

PR #187, opened 2023-01-16
try fixing widgets rendering

OriolAbril:
Adds some extra javascript to ensure widgets are rendered correctly
as static elements even without access to a python kernel.

Also does some minor doc config updates, mainly making the light theme the default.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,3 @@
+.jupyter-widgets-disconnected::before {
+  content: none !important;
+}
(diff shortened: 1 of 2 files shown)


OriolAbril:
looks basically the same now 🤔

aloctavodia:
Yep, I am not able  to spot any difference

OriolAbril:
Let's see how it looks now. It turns out ipywidgets adds one of these red broken chain icons _per element that should be connected to a kernel but isn't_. I guess it is informative, but a bit too much. I added an extra css rule to remove that.

aloctavodia:
Thanks, this looks much better!

PR #200, opened 2023-02-01
Add BetaBinomial distribution

aleicazatti:
## Description
WIP because pz.quartile function didn't work with this distribution, yet to find out why

## Checklist


preliz/distributions/discrete.py
```diff
@@ -116,6 +116,92 @@ def _fit_mle(self, sample):
...
+    ----------
+    n : int
+        Number of Bernoulli trials (n >= 0).
+    alpha : float
+        alpha > 0.
+    beta : float
+        beta > 0.
+    """
+
+    def __init__(self, alpha=None, beta=None, n=10):
```
aloctavodia, line 162:
```suggestion
    def __init__(self, alpha=None, beta=None, n=None):
```
`n` should be None, otherwise, for functions like maxent and quartile it will remain fixed

preliz/distributions/discrete.py
```diff
@@ -116,6 +116,92 @@ def _fit_mle(self, sample):
...
+    def _update(self, alpha, beta, n):
+        self.alpha = np.float64(alpha)
+        self.beta = np.float64(beta)
+        self.n = np.int64(n)
+        self.params = (self.alpha, self.beta, self.n)
+        self.support = (0, self.n)
+        self._update_rv_frozen()
+
+    def _fit_moments(self, mean, sigma):
+        # See https://en.wikipedia.org/wiki/Beta-binomial_distribution
```
aloctavodia, line 193:
```suggestion
        # See https://en.wikipedia.org/wiki/Beta-binomial_distribution#Method_of_moments
```

preliz/distributions/discrete.py
```diff
@@ -116,6 +116,92 @@ def _fit_mle(self, sample):
...
+        n = self.n
+        p = mean / n
```
aloctavodia, line 195: We can use the same crude approximation we use for the binomial

        n = mean + sigma * 2
        p = mean / n

preliz/distributions/discrete.py
```diff
@@ -116,6 +116,92 @@ def _fit_mle(self, sample):
...
+        rho = ((sigma**2 / (mean * (1 - p))) - 1) / (n - 1)
+        alpha = (p / rho) - p
+        beta = (alpha / p) - alpha
```
aloctavodia, line 198:
```suggestion
        alpha = max(0.5, (p / rho) - p)
        beta =  max(0.5, (alpha / p) - alpha)
```
the alpha and beta estimates can be negative. For those cases the value 0.5 is likely to be a better choice to be the starting point for the optimization than a very small positive value.

preliz/distributions/discrete.py
```diff
@@ -116,6 +116,94 @@ def _fit_mle(self, sample):
...
+        self.n = np.int64(n)
+        self.params = (self.alpha, self.beta, self.n)
+        self.support = (0, self.n)
+        self._update_rv_frozen()
+
+    def _fit_moments(self, mean, sigma):
+        # Crude aproximation for n (as in Binomial distribution)
+        # For alpha and beta see:
+        # https://en.wikipedia.org/wiki/Beta-binomial_distribution#Method_of_moments
+        n = np.int64(mean + sigma * 2)
```
aloctavodia, line 196:
```suggestion
        n = mean + sigma * 2
```
n will be turn into an integer by `_update`

aleicazatti: maybe if n is turned into an integer in this line alpha and beta are computed more accurately

aloctavodia: Yes, but we should not care about being too accurate about the moments. Better add this distribution to the list in this line https://github.com/arviz-devs/preliz/blob/1b50264cef0c64e9b5b542dfff40d7c4ec534d30/preliz/tests/test_distributions.py#L94

PR #212, opened 2023-02-14
add optimize_moments internal function

aloctavodia:
This PR adds an internal function to better approximate the moments, that we later use to initialize maxent. 


aloctavodia:
This may help you @aleicazatti with the moments/max-ent initialization of the hyper-geometric

PR #215, opened 2023-02-23
Add HyperGeometric distribution

aleicazatti:
## Description
Here's the problematic one @aloctavodia 

## Checklist

### DIFF
--- a/preliz/distributions/discrete.py
+++ b/preliz/distributions/discrete.py
@@ -604,9 +604,9 @@ class HyperGeometric(Discrete):
         self._update_rv_frozen()
 
     def _fit_moments(self, mean, sigma):
-        n = 2 * (mean + sigma * 2)
-        N = (mean * n * (n - mean)) / (n * (mean - sigma**2) - mean**2)
-        k = mean * N / n
+        n = mean + sigma * 4
+        k = n
+        N = k*n/mean
         params = N, k, n
         optimize_moments(self, mean, sigma, params)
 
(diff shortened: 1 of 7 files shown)


preliz/distributions/discrete.py
```diff
@@ -535,6 +536,84 @@ def _fit_mle(self, sample):
...
+        self.n = np.int64(n)
+        self.params = (self.N, self.k, self.n)
+        self.support = (max(0, n - N + k), min(k, n))
+        self._update_rv_frozen()
+
+    def _fit_moments(self, mean, sigma):
+        n = 2 * (mean + sigma * 2)
+        N = (mean * n * (n - mean)) / (n * (mean - sigma**2) - mean**2)
+        k = mean * N / n
+        params = N, k, n
```
aloctavodia, line 610:
```suggestion
        n = mean + sigma * 4
        k = n
        N = k*n/mean
        params = N, k, n
```

aloctavodia: I made this change and also 

`opt = least_squares(func, x0=init_vals, args=(dist, mean, sigma))`
to
`opt = minimize(func, x0=init_vals, args=(dist, mean, sigma), method="Powell")` 

in `optimize_moments`

with these changes maxent return results that look not that off. But there are still issues, mainly `optimize_max_ent` function returns exactly the same parameters used as the initial guess.

I did not check how the change in `optimize_moments` affects other distributions.

PR #217, opened 2023-02-24
use absolute loss for optimize_moments

aloctavodia:


aloctavodia:
Merging as I have rerun the test locally and on GitHub a few times and none of them showed a mismatch for the Rice distribution.

PR #218, opened 2023-02-24
Add Rice to assert exception list

aleicazatti:


aloctavodia:
I think this is not necessary given the change introduced in #217

PR #224, opened 2023-03-10
Add initial support for MultiVariate Variables

aloctavodia:
This adds general support for Multivariate variables and implements Dirichlet and Multivariate Normal in particular. Both implementations omit some methods compared to univariate. This is partly because of the differences between these two types of distributions, but also because we should think about which methods of elicitation we really want to implement for Multivariate distributions and work from there. One big missing method is `plot_interactive`, this should be added in the following PR:

### DIFF
--- a/preliz/distributions/continuous_multivariate.py
+++ b/preliz/distributions/continuous_multivariate.py
@@ -11,6 +11,7 @@ import numpy as np
 from scipy import stats
 
 from .distributions_multivariate import Continuous
+from .continuous import Beta, Normal
 from ..internal.distribution_helper import all_not_none
 from ..internal.plot_helper_multivariate import plot_dirichlet, plot_mvnormal
 
@@ -57,6 +58,7 @@ class Dirichlet(Continuous):
     def __init__(self, alpha=None):
         super().__init__()
         self.dist = copy(stats.dirichlet)
+        self.marginal = Beta
         self.support = (eps, 1 - eps)
         self._parametrization(alpha)
 
@@ -243,6 +245,7 @@ class MvNormal(Continuous):
     def __init__(self, mu=None, cov=None):
         super().__init__()
         self.dist = copy(stats.multivariate_normal)
+        self.marginal = Normal
         self.support = (-np.inf, np.inf)
         self._parametrization(mu, cov)
 
--- a/preliz/distributions/continuous_multivariate.py
+++ b/preliz/distributions/continuous_multivariate.py
@@ -96,16 +96,16 @@ class Dirichlet(Continuous):
     ):
         """
         Plot the pdf of the marginals or the joint pdf of the simplex.
-        The joint representation is only available for a dirichlet with a alpha of length 3.
+        The joint representation is only available for a dirichlet with an alpha of length 3.
 
         Parameters
         ----------
         marginals : True
-            Defaults to True, plor the marginal distributions, if False plot the joint distribution
+            Defaults to True, plot the marginal distributions, if False plot the joint distribution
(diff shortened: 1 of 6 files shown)


preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+        equal tailed interval `"eti"` or intervals defined by arbitrary `"quantiles"`.
+        levels : list
+            Mass of the intervals. For hdi or eti the number of elements should be 2 or 1.
+            For quantiles the number of elements should be 5, 3, 1 or 0
+            (in this last case nothing will be plotted).
+        support : str:
+            If ``full`` use the finite end-points to set the limits of the plot. For unbounded
+            end-points or if ``restricted`` use the 0.001 and 0.999 quantiles to set the limits.
+        figsize : tuple
+            Size of the
```
aleicazatti, line 120:
```suggestion
            Size of the figure
```

preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+        ax=None,
+    ):
+        """
+        Plot the cumulative distribution function.
+
+        Parameters
+        ----------
+        pointinterval : bool
+            Whether to include a plot of the quantiles. Defaults to False. If True the default is to
+            plot the median and two interquantiles ranges.
```
aleicazatti, line 148:
```suggestion
            plot the median and two interquantile ranges.
```

preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+        ax=None,
+    ):
+        """
+        Plot the quantile function.
+
+        Parameters
+        ----------
+        pointinterval : bool
+            Whether to include a plot of the quantiles. Defaults to False. If True the default is to
+            plot the median and two interquantiles ranges.
```
aleicazatti, line 187:
```suggestion
            plot the median and two interquantile ranges.
```

preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+
+    ========  ==========================
+    Support   :math:`x \in \mathbb{R}^k`
+    Mean      :math:`\mu`
+    Variance  :math:`T^{-1}`
+    ========  ==========================
+
+    Parameters
+    ----------
+    mu : tensor_like of float
```
aleicazatti, line 239:
```suggestion
    mu : array of floats
```

preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+        Vector of means.
+    cov : tensor_like of float, optional
```
aleicazatti, line 241:
```suggestion
    cov : array of floats, optional
```

preliz/distributions/continuous_multivariate.py
```diff
@@ -0,0 +1,396 @@
...
+        ax=None,
+    ):
+        """
+        Plot the  pdf of the marginals or the joint pdf
+        The joint representation is only available for a 2D Multivariate Normal.
+
+        Parameters
+        ----------
+        marginals : True
+            Defaults to True, plor the marginal distributions, if False plot the joint distribution
```
aleicazatti, line 294:
```suggestion
            Defaults to True, plot the marginal distributions, if False plot the joint distribution
```

aleicazatti (commented):
Suggestions for some typos and small changes in docstrings.

PR #259, opened 2023-06-14
Add dependency bundles

### LINKED ISSUE #255: `notebook` dependency
(I'm a reviewer for the submission of `preliz` at JOSS)

This is in relation to the review criterion "Installation instructions: Is there a clearly-stated list of dependencies?".

Given the likely use of a jupyter notebook for the interactive aspects of the package, I'm wondering if `notebook` might be best listed as a dependency? Alternatively, perhaps it could be flagged in the "Installation" section of the docs that a `pip install notebook` may be required and again when `plot_interactive` is introduced.

aleicazatti:
## Description

* Add dependency bundles
* Update README and index
* Close  #255

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,4 +13,4 @@ python:
    - method: pip
      path: .
      extra_requirements:
-       - doc
+       - full
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,12 +3,14 @@
 version: 2
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-22.04
   tools:
-    python: "3.9"
+    python: "3.10"
 
 python:
    install:
    - requirements: requirements-docs.txt
    - method: pip
      path: .
+     extra_requirements:
+       - doc
--- a/README.md
+++ b/README.md
@@ -19,28 +19,24 @@ The PreliZ documentation can be found in the [official docs](https://preliz.read
 
 ### Last release
 PreliZ is available for installation from [PyPI](https://pypi.org/project/preliz/).
-There are two dependency bundles for the latest version:
+The latest version (base set of dependencies) can be installed using pip:
 
-* `full`: includes dependencies for all features (interactive and non-interactive)
-* `non-interactive`: includes dependencies for non-interactive features only
+```
+pip install preliz
+```
+To make use of the interactive features, you can install the optional dependencies:
 
-And both can be installed using pip:
+* For JupyterLab:
 
-::::{tab-set}
-:::{tab-item} Full
-
-```bash
-pip install "preliz[full]"
+```
+pip install "preliz[full,lab]"
 ```
 
-:::
-:::{tab-item} Non-interactive
+* For Jupyter Notebook:
 
-```bash
-pip install "preliz[non-interactive]"
 ```
-:::
-::::
+pip install "preliz[full,notebook]"
+```
 
 ### Development
 The latest development version can be installed from the main branch using pip:
--- a/README.md
+++ b/README.md
@@ -19,11 +19,28 @@ The PreliZ documentation can be found in the [official docs](https://preliz.read
 
 ### Last release
 PreliZ is available for installation from [PyPI](https://pypi.org/project/preliz/).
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -19,11 +19,28 @@ The PreliZ documentation can be found in the [official docs](https://preliz.read
...
+```bash
+pip install "preliz[full]"
 ```
-pip install preliz
+
+:::
+:::{tab-item} Non-interactive
+
+```bash
+pip install "preliz[non-interactive]"
```
OriolAbril, line 40: This one should be `preliz`. The base set of dependencies (the ones specified [here](https://github.com/arviz-devs/preliz/blob/main/pyproject.toml#L25-L33)) are always installed. We can then also skip the `non-interactive` bundle.

README.md
```diff
@@ -19,11 +19,28 @@ The PreliZ documentation can be found in the [official docs](https://preliz.read
...
 PreliZ is available for installation from [PyPI](https://pypi.org/project/preliz/).
-The latest  version can be installed using pip:
+There are two dependency bundles for the latest version:
 
+* `full`: includes dependencies for all features (interactive and non-interactive)
+* `non-interactive`: includes dependencies for non-interactive features only
+
+And both can be installed using pip:
+
+::::{tab-set}
```
OriolAbril, line 29: tabs do not work in the github readme nor on pypi, so I would not use them here, only in the docs (there is can be used from rst, md or ipynb files, but it is a sphinx feature)

pyproject.toml
```diff
@@ -39,7 +39,23 @@ name = "preliz"
...
+  "ipywidgets",
+  "ipympl",
+  "jupyterlab",
+  "notebook",
+]
+non-interactive = [
+  "arviz",
+  "numpy>=1.22",
+  "scipy>=1.9.1",
+]
```
OriolAbril, line 58: Assuming we update the dependencies section (see comment above) to:

```
dependencies = [
  "arviz",
  "numpy>=1.22",
  "scipy>=1.9.1",
]
```

I would then update that to one of:

**opinionated version**: use jupyterlab

```
[project.optional-dependencies]
full = [
  "matplotlib>=3.5",
  "nbclient<0.6,>=0.2",
  "ipywidgets",
  "ipympl",
  "jupyterlab",
]
```

**ide agnostic version**:

```
[project.optional-dependencies]
full = [
  "matplotlib>=3.5",
  "nbclient<0.6,>=0.2",
  "ipywidgets",
  "ipympl",
]
lab = ["jupyterlab"]
notebook = ["notebook"]
```

which would modify install commands to:

```shell
pip install preliz # base
pip install "preliz[full,lab]"  # jupyterlab
pip install "preliz[full,notebook]"  # jupyter notebook
```

OriolAbril (commented):
left some comments about how exactly we want to bundle the dependencies and optional tags. The current approach is a bit inconsistent. Later on we can discuss doc specifics, wordings and such

OriolAbril:
I hope this will fix tests and docs, but we should probably set up a github action for testing "base" dependencies. As of now, the package might not be importable unless we do `preliz[full]`

.readthedocs.yaml
```diff
@@ -3,12 +3,14 @@
...
-    python: "3.9"
+    python: "3.10"
 
 python:
    install:
    - requirements: requirements-docs.txt
    - method: pip
      path: .
+     extra_requirements:
+       - doc
```
OriolAbril, line 16:
```suggestion
       - full
```

aloctavodia (commented):
LGTM

pyproject.toml
```diff
@@ -39,7 +35,15 @@ name = "preliz"
 source = "https://github.com/arviz-devs/preliz"
 tracker = "https://github.com/arviz-devs/preliz/issues"
 
-
+[project.optional-dependencies]
+full = [
+  "matplotlib>=3.5",
```
aloctavodia, line 40: Shouldn't matplotlib be listed as part of the regular dependencies? Also currently matplotlib will be installed anyway as we have arviz as a dependency

OriolAbril: Yes

PR #309, opened 2024-02-03
Add elicitation of Beta distribution with bounds and mode

nishant42491:
## Checklist

- [ ] Includes new or updated tests to cover the new feature


demo.py
```diff
@@ -0,0 +1,94 @@
...
+    alpha2 = 1 + (1 - mode) * tau_bisected
+
+    cur_prob = calc_cdf([l1, l2], aplha1, alpha2)
+
+    print("tau_star: ", tau_star)
+    print("tau_bisected: ", tau_bisected)
+    print("alpha1: ", aplha1)
+    print("alpha2: ", alpha2)
+    print("cur_prob: ", cur_prob)
+
```
aloctavodia, line 94:
```suggestion
import preliz as pz

def calc_cdf(l1, u1, a, b):
    dist = pz.Beta(a, b)
    return dist.cdf(u1) - dist.cdf(l1)

def one_iter(l1, u1, stop_prob=0.99, eps=0.001):

    tau_not = 0
    mode = (l1 + u1) / 2

    prob = calc_cdf(l1, u1, 1, 1)
    
    while abs(prob - stop_prob) > eps:

        tau_not +=  0.1

        alpha1 = 1 + mode * tau_not
        alpha2 = 1 + (1 - mode) * tau_not

        prob =  calc_cdf(l1, u1, alpha1, alpha2)

    return alpha1, alpha2, prob



l1 = 0.25
l2 = 0.75
mode = (l1 + l2) / 2
alpha1, alpha2, cur_prob = one_iter(l1, l2)


print("alpha1: ", alpha1)
print("alpha2: ", alpha2)
print("cur_prob: ", cur_prob)
pz.Beta(alpha1, alpha2).plot_pdf()
pz.maxent(pz.Beta(), l1, l2, 0.99);
```

aloctavodia: I was not able to run your code for some combinations of l1,  l2, so I simplified and rewrote a few things. Feel free to add some of the deleted functions if you need them. 

I think this is doing what the paper describes, tell me if I made a mistake. 

As you can see I add a comparison to the maxent function. You will see that both tend to agree, at least for a certain range of l1, and l2 inputs. You will see the agreement improves when the beta becomes more symmetrical (this is when alpha1 and alpha2 become closer), the reason is that these methods assume the mode is in the middle of the interval. Instead, the maxent function returns the maximum entropy distribution.

aloctavodia (changes requested):
There are a few files under the .idea folder that should not be part of this PR

aloctavodia:
As you can see for the Beta case, the maxent function already does something similar and in my opinion in a much more flexible way. Anyway, it may be the case that someone prefers this method. In that case, this function should have the l1, l2 and mode as input values and it should return a pz.Beta distribution with the fitted parameter and optionally plot the result. Probably the eps can be fixed as 0.005.

Check pz.maxent for a reference of parameters names and other details

aloctavodia:
Thanks!

I made a few changes. Check `maxent` as a reference for further changes, like adding a docstring and refactoring. Notice that all optimization routines have to be at `interval/optimization.py`

The main changes are that the signature is (lower, upper, mode, mass=0.9, plot=True, plot_kwargs=None, ax=None), meaning the mode is a mandatory argument, together with the lower and upper bounds.

Also, notice that it returns a frozen pz.Beta() distribution, meaning we have already passed the parameters. 

To compute the cdf I did a little trick. First I call pz.Beta(.,.) but then I use the `_parametrization`, this is slightly faster. Instantiating a distribution in PreliZ (and SciPy) can be relatively costly.

aloctavodia:
BTW, notice that we are slightly departing here from what is is presented in the paper in the sense that we are demanding the user to provide a mode and allowing them to select the mass (instead of fixing it at 0.99).

Ohh another think still missing is that we should check that the mode is between lower and upper, otherwise raise an error

nishant42491:
> Thanks!
> 
> I made a few changes. Check `maxent` as a reference for further changes, like adding a docstring and refactoring. Notice that all optimization routines have to be at `interval/optimization.py`
> 
> The main changes are that the signature is (lower, upper, mode, mass=0.9, plot=True, plot_kwargs=None, ax=None), meaning the mode is a mandatory argument, together with the lower and upper bounds.
> 
> Also, notice that it returns a frozen pz.Beta() distribution, meaning we have already passed the parameters.
> 
> To compute the cdf I did a little trick. First I call pz.Beta(.,.) but then I use the `_parametrization`, this is slightly faster. Instantiating a distribution in PreliZ (and SciPy) can be relatively costly.

Thanks for making all the changes. I'll check the maxent reference to update all the docstrings for it

nishant42491:
> BTW, notice that we are slightly departing here from what is is presented in the paper in the sense that we are demanding the user to provide a mode and allowing them to select the mass (instead of fixing it at 0.99).
> 
> Ohh another think still missing is that we should check that the mode is between lower and upper, otherwise raise an error

should i include an assert statement to the check the bounds of the mode?

demo.py
nishant42491, line 1: I have added an extra argument eps for the tolerance parameter instead of setting it to 0.005 and I have added docstrings too

nishant42491:
I have added docstrings and have tried refactoring the code to follow the standards set in the maxent function. I have also added an extra parameter eps instead of hard setting the tolerance level to 0.005.

nishant42491:
I'm not sure what optimisations are exactly required for the function. Also where would this module exist inside of the preliz structure? maybe somewhere in the pz.unidimentional directory and what should be the name of the module and the function one_iter exactly?

demo.py
```diff
@@ -0,0 +1,85 @@
...
+    if not 0 < mass <= 1:
+        raise ValueError("mass should be larger than 0 and smaller or equal to 1")
+
+    if upper <= lower:
+        raise ValueError("upper should be larger than lower")
+
+    alpha = 1
+    beta = 1
+    dist = pz.Beta(alpha, beta)
```
aloctavodia, line 50: probably we can do better than this based on user-passed values, as we do in maxent. Something like 

```python
    dist = pz.Beta()
    dist._fit_moments(  # pylint:disable=protected-access
        mean=(upper+lower)/2, 
        sigma=((upper - lower) / 6) / mass
    )
    tau_a = (dist.alpha - 1) / mode
    tau_b = (dist.beta - 1) / (1-mode)
    tau = (tau_a + tau_b) / 2
```
this should be a better guessing than tau=0. Maybe we can do something even better not sure yet.

then inside of the while loop, we should allow tau to increase or decrease (in case our initial guessing is higher than it should be). Something like

```python
        if prob < mass:
            tau += 0.1
        else:
            tau -= 0.1
```
On average this guessing should make the algorithm faster

aloctavodia (changes requested):
thanks, a few more comments. Also we need to think about a name for the function

nishant42491:
@aloctavodia  thanks for all your ideas. I have tried to clean up the code by following your comments. I cannot think of a function name though the closest thing I can think of is beta_bounds_elicitation. I have added an optimisation routine let me know if that is what you were looking for.  additionally, I have tried to make the plotting routine resemble the one in the maxent function however when we keep the default argument as None for the plot kwargs arg will raise an error as it expects it to be a mapping.
I have also added the tau calculation routine which you had suggested to make our function run faster.

aloctavodia:
A test is needed. We can add it as part of this PR, or in a different one.

aloctavodia (approved):

nishant42491:
@aloctavodia I think maybe we can let the tests be a separate PR.  Do you need anything else to be done as part of this PR?

aloctavodia:
Thanks @nishant42491!

Issue #127, opened 2022-12-10
Improve ipywidgets layout for roulette


aloctavodia:
* Improve aesthetics 
* Separate the list of distributions into discrete and continuous. Or create a panel of checkboxes with general control like "all discrete", "all continuous", "deselect all"... This panel could be placed to the right of the matplotlib subplot to take advantage of the blank space (and wide screens).

rohanbabbar04:
@aloctavodia I was thinking about this issue day before yesterday.

Would the second point be something like this

![image](https://github.com/arviz-devs/preliz/assets/77425744/bf887966-91e2-4bc0-8df2-94da4dee3f0d)

aloctavodia:
maybe, Not sure about custom, what will happen when the users click "custom"? 
I guess that if the user selects the distributions from the dropdown menu then "custom" will turn on. Right?

What about having 3 checkboxes, Continuous, Discrete, and None? 

*  It should be possible to click on Continuous, and Discrete at the same time (to select all). 
* It should not be possible to click on None and something else. 
* Deselecting Continous and Discrete has the same result as clicking None
* Selecting None cleans all selections, this is useful to clean user custom selections.

rohanbabbar04:
Yep, that's what `custom` would do...

I like the idea of having Continuous, Discrete, and None(easier to understand). Let's go with that. Is that fine?

aloctavodia:
Yes, that's fine.

rohanbabbar04:
With #318  merged @aloctavodia do you want me to do anything on the aesthetics...
Also we should open an issue for updating the `.gif`

aloctavodia:
I don't have any clear any about the asthetics. Yes we need to update the gif.

PR #318, opened 2024-02-14
Add checkboxes to control distributions in roulette

rohanbabbar04:
Solves #127 
## Description
Added checkboxes to control the roulette function

- Added three checkboxes `Continuous`, `Discrete`, and `None`.
- Added `handle_checkbox_change` as a handler function.

An image
image

## Checklist


rohanbabbar04:
- Have kept the control of checkboxes separated
- At times, it is a tad too slow
- Once this is approved, we may also have to change `roulette.gif`

aloctavodia:
What is slow? If the user pass `dist_names` interacting with the checkbox will

rohanbabbar04:
Maybe....
I think ipywidgets in interactive mode are also sometimes slow...

PR #322, opened 2024-02-18
Update roulette plot and gif

rohanbabbar04:
## Description
- Updated `roulette.gif` after considering the changes of #318 
- Saved the `roulette` plot in Direct Elicitation in 1D.

## Checklist

- [x] Update the gif for interactive roulette function


rohanbabbar04:
I think the current ipywidgets look good and fulfills the functionality of the roulette plot.
What do you think?
We can maybe close the issue, and if required we can reopen/create a more specific issue in the future...

aloctavodia (approved):

PR #323, opened 2024-02-18
added beta mode tests

### LINKED ISSUE #320: Add test beta_mode
see #309 for details

### LINKED ISSUE #325: beta_mode: plot should default to True
To be consistent with other functions

nishant42491:
## Description

This PR adds tests for beta_mode and makes changes to the beta call function interface.
resolves #325,  resolves #320 

## Checklist

- [X] Code style is correct (follows pylint and black guidelines)
- [X] Includes new or updated tests to cover the new feature

### DIFF
--- a/preliz/tests/test_beta_mode.py
+++ b/preliz/tests/test_beta_mode.py
@@ -6,7 +6,7 @@ from preliz import beta_mode
 @pytest.mark.parametrize(
     "lower, upper, mode, mass, expected_alpha, expected_beta",
     [
-        (0, 1, 0.5, 0.9, 2.184, 2.184),  # Example test case
+        (0.25, 0.75, 0.5, 0.9, 2.184, 2.184),  # Example test case
         # Add more test cases here
     ],
 )
@@ -18,25 +18,22 @@ def test_beta_mode(lower, upper, mode, mass, expected_alpha, expected_beta):
     assert np.isclose(dist.beta, expected_beta, atol=0.01)
 
 
-# Additional tests for input validation
-
 def test_invalid_mass():
     with pytest.raises(ValueError):
-        beta_mode(0, 1, 0.5, mass=1.1)
+        beta_mode(0.25, 0.75, 0.5, mass=1.1)
 
 
 def test_invalid_mode():
     with pytest.raises(ValueError):
-        beta_mode(0, 1, 1.5)
+        beta_mode(0.25, 0.75, 1.5)
 
 
 def test_invalid_bounds():
     with pytest.raises(ValueError):
-        beta_mode(1, 0, 0.5)
+        beta_mode(0.75, 0.25, 0.5)
 
-#additional tests for plot
 
 def test_plot_beta_mode():
     _, dist = beta_mode(0, 1, 0.5, 0.9, plot=True,
                         plot_kwargs={"pointinterval": True})
-    # Add more assertions as needed
+
--- a/preliz/tests/test_beta_mode.py
+++ b/preliz/tests/test_beta_mode.py
@@ -33,5 +33,4 @@ def test_invalid_bounds():
 
 
 def test_plot_beta_mode():
-    _, _ = beta_mode(0.25, 0.75, 0.5, 0.9, plot=True,
-                        plot_kwargs={"pointinterval": True})
+    _, _ = beta_mode(0.25, 0.75, 0.5, 0.9, plot=True, plot_kwargs={"pointinterval": True})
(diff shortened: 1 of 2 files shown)


nishant42491:
@aloctavodia  is this what you were broadly looking for?

aloctavodia:
The beta distribution is defined on the 0-1 interval. So for any beta distribution 100% of the mass will necessarily be inside that interval. In other words it is not posible to define a beta distribution with a mass less than 1 if the lower bound is 0 and the upper bound is 1.

aloctavodia:
Hi @nishant42491 Let me know if you need help.

nishant42491:
> Hi @nishant42491 Let me know if you need help.

hey there, I was a bit confused about how to compare the values of the fitted distribution. maybe use the distribution parameters which we receive from maxent let me know what you think. 

Also, I have included a minor fix to the mode value error included in the beta_mode function. The function does not converge when mode = lower bound or mode = upper bound hence a strict inequality is needed for that case.

nishant42491:
> The beta distribution is defined on the 0-1 interval. So for any beta distribution 100% of the mass will necessarily be inside that interval. In other words it is not posible to define a beta distribution with a mass less than 1 if the lower bound is 0 and the upper bound is 1.

i have tried to fix this in my latest commit.

preliz/tests/test_beta_mode.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+import numpy as np
+from preliz import beta_mode
+
+
+@pytest.mark.parametrize(
+    "lower, upper, mode, mass, expected_alpha, expected_beta",
+    [
+        (0.25, 0.75, 0.5, 0.9, 6.41, 9.12),  # Example test case
```
aloctavodia, line 9: the value of alpha and beta seems to be right. this should be a symmetrical distribution and hence alpha and beta should be equal.

preliz/tests/test_beta_mode.py
```diff
@@ -0,0 +1,39 @@
...
+        beta_mode(0.25, 0.75, 1.5)
+
+
+def test_invalid_bounds():
+    with pytest.raises(ValueError):
+        beta_mode(0.75, 0.25, 0.5)
+
+
+def test_plot_beta_mode():
+    _, dist = beta_mode(0, 1, 0.5, 0.9, plot=True,
```
aloctavodia, line 37: the lower and upper bounds are still incorrect. 

Maybe we can unify this test and the first one

nishant42491:
yea, you are right the dist parameters should be symmetric. I have added a fix for that.

nishant42491:
yep, black seems to have fixed all the linting issues

preliz/tests/test_beta_mode.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+import numpy as np
+from preliz import beta_mode
+
+
+@pytest.mark.parametrize(
+    "lower, upper, mode, mass, expected_alpha, expected_beta",
+    [
+        (0.25, 0.75, 0.5, 0.9, 4.94, 4.94),  # Example test case
+        # Add more test cases here
```
aloctavodia, line 10: remove these comments. 

Are you going to add more tests here? If not there is not need to use "@pytest.mark.parametrize" just write the values inside the function

nishant42491: i have removed the comments and pytest parameterise in the latest commit.

nishant42491:
I have also set the default plot keyword to True in Beta_Mode which should fix issue #325. I have linked the issue through the description of this PR.

preliz/tests/test_beta_mode.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+import numpy as np
+from preliz import beta_mode
+
+
+def test_beta_mode():
+
+    test_case = (0.25, 0.75, 0.5, 0.9, 4.94, 4.94)
```
aloctavodia, line 8: No need to define a tuple just to unpack it on the next line

aloctavodia (changes requested):

preliz/tests/test_beta_mode.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+import numpy as np
+from preliz import beta_mode
+
+
+def test_beta_mode():
+
+    lower, upper, mode, mass, expected_alpha, expected_beta = 0.25, 0.75, 0.5, 0.9, 4.94, 4.94
```
aloctavodia, line 8: Just pass the values to the functions. Don't need to define this variables

nishant42491: oh okay my bad. i have fixed it

aloctavodia (approved):

PR #327, opened 2024-02-21
Dirichlet Elicitation [WIP]

### LINKED ISSUE #202: Add method to elicit a Dirichlet prior
There are a few methods as described for example [here](https://arxiv.org/abs/2112.01380). 

This [method](https://jeremy-oakley.shinyapps.io/SHELF-Dirichlet/) or this  [one](https://www.mdpi.com/1099-4300/19/10/564) may be good candidates

nishant42491:
## Description

fixes #202 

## Checklist

- [X] Code style is correct (follows pylint and black guidelines)
- [X] Includes new or updated tests to cover the new feature
- [X] New features are properly documented (with an example if appropriate)


nishant42491:
The sample case in the demo corresponds to Example 3 in this paper https://www.mdpi.com/1099-4300/19/10/564. The values are an exact match so the implementation seems to be correct.

aloctavodia:
thanks @nishant42491 please check the changes and let me know if you have questions or doubts

aloctavodia:
I have been playing with the implementation and re-reading the paper. And still find the elicitation procedure unintuitive. Hence if you check my last commit you will find an alternative method. The user provides a mode, a mass, and a "bound". The method returns the Dirichlet distribution with the marginals having that mass around the mode +- the bound.  This is an approximate solution, the accuracy increases when the marginals are more "gaussian-like". 

I find this approach more appealing.  What do you think?

nishant42491:
> I have been playing with the implementation and re-reading the paper. And still find the elicitation procedure unintuitive. Hence if you check my last commit you will find an alternative method. The user provides a mode, a mass, and a "bound". The method returns the Dirichlet distribution with the marginals having that mass around the mode +- the bound. This is an approximate solution, the accuracy increases when the marginals are more "gaussian-like".
> 
> I find this approach more appealing. What do you think?

Yep, this one feels more intuitive to me as well. To be honest, the simplex approach described in the paper was very unintuitive for me too.  

I have a question how do we make the marginals of the Dirichlet more gaussain-like?

aloctavodia:
Notice that we still have a simplex, as the Dirichlet is defined on the simplex and internally the approach is very similar. The main difference are that we define the lower bounds as a user-defined deviation from the mode, we don't use a MC approximation, instead we evaluate the cdf of the marginals. And we allow the user to define a mass, so they can control the level of uncertainty. 

By gaussian-like I meant the distribution is close to be symmetrical, this happens when the marginal gets concentrated and/or away from the boundaries.

aloctavodia:
OK, do you want to turn this onto a proper Preliz function. I guess we can call it dirichlet_mode as it similar to the beta_mode function

nishant42491:
> OK, do you want to turn this onto a proper Preliz function. I guess we can call it dirichlet_mode as it similar to the beta_mode function

yep i have started working on the proper implementation.

nishant42491:
@aloctavodia I have created an optimisation routine and skeleton for the main function here. Do let me know if you want any changes to those. Also plotting the Dirichlet distribution through the plot_pdf returns a None I don't know why.

aloctavodia:
https://github.com/arviz-devs/preliz/blob/main/preliz/internal/plot_helper_multivariate.py#L88-L173 is returning None.

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
+from preliz.distributions import Dirichlet, Beta
+import preliz as pz
+import numpy as np
+from preliz.internal.optimization import optimize_dirichlet_mode
+import logging
```
aloctavodia, line 5: This should be the first import

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
...
+
+    alpha = [1 + tau * mode_i for mode_i in mode]
+
+    a_0 = sum(alpha)
+    marginal_prob_list = []
+    for a_i, lbi in zip(alpha, lower_bounds):
+        _dist._parametrization(a_i, a_0 - a_i)
+        marginal_prob_list.append(_dist.cdf(lbi))
+    mean_cdf = np.mean(marginal_prob_list)
+    return mean_cdf, alpha
```
aloctavodia, line 20: remove this

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def dirichlet_mode(mode, mass=0.90, bound=0.01, plot=True, plot_kwargs={}, ax=None):
+    """
+
+    This function returns a Dirichlet distribution that has the specified mass concentrated in the region of
+    mode +- bound.
```
aloctavodia, line 27:
```suggestion
   This function returns a Dirichlet distribution that has the specified mass concentrated in the
   region of mode +- bound.
```

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Parameters
+    ----------
+
+    mode : list
+        Mode of the Dirichlet distribution.
+
```
aloctavodia, line 34: remove blank lines

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    if not 0 < mass <= 1:
+        raise ValueError("mass should be larger than 0 and smaller or equal to 1")
+
+    if not all(i > 0 for i in mode):
+        raise ValueError("mode should be larger than 0")
+
+    if not abs(sum(mode) - 1) < 0.0001:
+        _log.warning("The mode should sum to 1, normalising mode to sum to 1")
```
aloctavodia, line 70: Use warnings. The message should return the normalized mode, so the user can check if they are happy with it.

```suggestion
        warnings.warn("The mode should sum to 1, normalising mode to sum to 1")
```

dirichlet_prior_demo.py
```diff
@@ -0,0 +1,111 @@
...
+
+mode = [0.4, 0.3, 0.2, 0.1]
+mass = 0.90
+bound = 0.01
+
+Dirichlet_dist = dirichlet_mode(mode, mass, bound)
+
+alpha = Dirichlet_dist[1].alpha
+mode = (alpha - 1) / (alpha.sum() - len(alpha))
+print("Mode", mode)
```
aloctavodia, line 111:
```suggestion
```

preliz/internal/optimization.py
```diff
@@ -186,6 +186,32 @@ def negll(params, dist, sample):
     return opt
 
 
+def optimize_dirichlet_mode(lower_bounds, mode, tau, new_prob, target_mass, _dist, alpha):
```
aloctavodia, line 189: This is missing the docstring, that is present on `dirichlet_prior_demo.py`. Move the relevant information here and deleted `dirichlet_prior_demo.py`.

aloctavodia (changes requested):

preliz/internal/optimization.py
nishant42491: this is about the prob_approx function, right? ordo you mean something else?

aloctavodia: Ohh no, I just misread the file

nishant42491:
@aloctavodia  I have made the necessary changes can you please review them thank you .

aloctavodia (commented):
No need to tag me. I get notified when you submit changes. 

This is still not in the proper form. The name of the function should not have "demo", you need to update the __init__.py so you can call this from preliz. Tests are missing.

aloctavodia (changes requested):
Please, fix the linter issues and ensure test pass

preliz/unidimensional/dirichlet_mode.py
```diff
@@ -0,0 +1,75 @@
...
+import numpy as np
+import warnings
+
+_log = logging.getLogger("preliz")
+
+
+def dirichlet_mode(mode, mass=0.90, bound=0.01, plot=True, plot_kwargs=None, ax=None):
+    """
+    This function returns a Dirichlet distribution that has the specified mass concentrated in the
+    region of mode +- bound.
```
aloctavodia, line 13:
```suggestion
    Returns a Dirichlet distribution where the marginals have the specified mode and mass and their masses lie within the range mode ± bound. 
```

preliz/unidimensional/dirichlet_mode.py
```diff
@@ -0,0 +1,75 @@
...
+
+     Parameters
+     ----------
+     mode : list
+         Mode of the Dirichlet distribution.
+     mass : float
+         Probability mass between ``lower`` and ``upper`` bounds. Defaults to 0.90.
```
aloctavodia, line 20:
```suggestion
         Probability mass between within mode +- bounds. Defaults to 0.90.
```

preliz/unidimensional/dirichlet_mode.py
```diff
@@ -0,0 +1,75 @@
...
+     bound : float
+         Defines upper and lower bounds for the mass as mode+-bound. Defaults to 0.01.
```
aloctavodia, line 22:
```suggestion
         Defines upper and lower bounds for the mass as mode +- bound. Defaults to 0.01.
```

preliz/unidimensional/dirichlet_mode.py
```diff
@@ -0,0 +1,75 @@
...
+    lower_bounds = np.clip(np.array(mode) - bound, 0, 1)
+    target_mass = (1 - mass) / 2
+    _dist = Beta()
+
+    new_prob, alpha = optimize_dirichlet_mode(lower_bounds, mode, target_mass, _dist)
+
+    alpha_np = np.array(alpha)
+    calculated_mode = (alpha_np - 1) / (alpha_np.sum() - len(alpha_np))
+
+    if np.linalg.norm(np.array(mode) - calculated_mode) > 0.01:
```
aloctavodia, line 63: This could be something like this

```suggestion
    if np.any((np.array(mode) - calculated_mode) > 0.01):
```

aloctavodia:
You need to run black on preliz/unidimensional/dirichlet_mode.py

nishant42491:
> You need to run black on preliz/unidimensional/dirichlet_mode.py

okay ill do that now

nishant42491:
i guess all tests are passing now

aloctavodia (approved):

PR #356, opened 2024-03-15
Add faster Laplace

rohanbabbar04:
## Description
- Add class `Laplace` with all methods.
- Update test in `test_scipy.py`.

## Checklist

- [x] New features are properly documented (with an example if required)

### DIFF
--- a/preliz/distributions/laplace.py
+++ b/preliz/distributions/laplace.py
@@ -69,7 +69,7 @@ class Laplace(Continuous):
         Compute the probability density function (PDF) at a given point x.
         """
         x = np.asarray(x)
-        return nb_pdf(x, self.mu, self.b)
+        return np.exp(nb_logpdf(x, self.mu, self.b))
 
     def cdf(self, x):
         """
@@ -89,7 +89,7 @@ class Laplace(Continuous):
         """
         Compute the log probability density function (log PDF) at a given point x.
         """
-        return np.log(nb_pdf(x, self.mu, self.b))
+        return nb_logpdf(x, self.mu, self.b)
 
     def _neg_logpdf(self, x):
         """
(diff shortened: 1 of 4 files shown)


preliz/distributions/laplace.py
```diff
@@ -0,0 +1,171 @@
...
+        Compute the percent point function (PPF) at a given probability q.
+        """
+        q = np.asarray(q)
+        return nb_ppf(q, self.mu, self.b)
+
+    def logpdf(self, x):
+        """
+        Compute the log probability density function (log PDF) at a given point x.
+        """
+        return np.log(nb_pdf(x, self.mu, self.b))
```
aloctavodia, line 92: This should be the other way around, write an explicit expression for logpdf, and exponentiate that to compute the pdf.

preliz/distributions/laplace.py
```diff
@@ -0,0 +1,171 @@
...
+    if q > 0.5:
+        q = -np.log(2 * (1 - q))
+    else:
+        q = np.log(2 * q)
+    return q * b + mu
+
+
+@nb.njit
+def nb_neg_logpdf(x, mu, b):
+    return (-np.log(nb_pdf(x, mu, b))).sum()
```
aloctavodia, line 159: define nb_logpdf and use it

aloctavodia (changes requested):

preliz/distributions/laplace.py
rohanbabbar04: Noted, will keep that in mind....

preliz/distributions/laplace.py
rohanbabbar04: Done

aloctavodia (approved):

PR #364, opened 2024-03-19
Add faster Asymmetric Laplace

rohanbabbar04:
## Description
- Add class `AsymmetricLaplace` with all methods.
- Update test in `test_scipy.py`.

## Checklist


rohanbabbar04:
- I was getting an `OverFlow Error` for the line which I commented out having `ZeroInflatedBinomial`.

For the `rvs` function, I implemented https://en.m.wikipedia.org/wiki/Asymmetric_Laplace_distribution#Generating_asymmetric_Laplace_variates, but the results do not match...
If you have time can you also try it and let me know, also it works only for a specific range (-κ, 1/κ).

Currently, I have set it to the `self.ppf`(which is the other obvious alternative).

aloctavodia:
Can you open an issue for the overflow of betabinomial?
Can you send me the code for the rvs that do not match (here or by e-mail)? 

You are right using the inverse cdf method is a general method to obtain random samples. One reason to have custom samples is speed. But maybe here it will be enough.

rohanbabbar04:
> Can you open an issue for the overflow of zeroinflatedbinomial? Can you send me the code for the rvs that do not match (here or by e-mail)?
> 
> You are right using the inverse cdf method is a general method to obtain random samples. One reason to have custom samples is speed. But maybe here it will be enough.

- Sure, I will open an issue for the `OverflowError`.
- If using self.ppf is enough to handle the `rvs` you can review the PR, I will send you the code for the `rvs` mentioned in the link.

aloctavodia:
I will wait for the rvs code before reviewing

rohanbabbar04:
> I will wait for the rvs code before reviewing

Sure, currently a little busy will provide you with the code tomorrow.

preliz/distributions/asymmetric_laplace.py
```diff
@@ -0,0 +1,238 @@
...
+        return 2.0 * (1 - np.power(self.kappa, 6)) / np.power(1 + np.power(self.kappa, 4), 1.5)
+
+    def kurtosis(self):
+        return 6.0 * (1 + np.power(self.kappa, 8)) / np.power(1 + np.power(self.kappa, 4), 2)
+
+    def rvs(self, size=1, random_state=None):
+        random_state = np.random.default_rng(random_state)
+        random_samples = random_state.uniform(size=size)
+        # Returning ppf of random samples
+        return self.ppf(random_samples)
```
aloctavodia, line 175: something like this should work

@nb.njit()
def arvs(mu, b, kappa, size=1):
    u = np.random.uniform(-kappa, 1/kappa, size)
    s = np.sign(u)
    return mu - (1 / (1/b * s * kappa**s)) * np.log(1 - u * s * kappa**s)

preliz/distributions/asymmetric_laplace.py
```diff
@@ -0,0 +1,238 @@
...
+        if self.kappa > 1:
+            return self.mu + self.kappa * self.b * np.log(
+                (1 + self.kappa**2) / (2 * self.kappa**2)
+            )
+        return self.mu - np.log((1 + self.kappa**2) / 2) / (self.kappa / self.b)
+
+    def mean(self):
+        kap_inv = 1 / self.kappa
+        kappa_dif = kap_inv - self.kappa
+        return kappa_dif * self.b + self.mu
```
aloctavodia, line 156: rewrite in a single line

preliz/distributions/asymmetric_laplace.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def var(self):
+        kap_inv = 1 / self.kappa
+        return (kap_inv**2 + self.kappa**2) * self.b**2
```
aloctavodia, line 160: rewrite in a single line

aloctavodia (changes requested):

preliz/distributions/asymmetric_laplace.py
rohanbabbar04: Almost similar I did something like this
```
@nb.vectorize(nopython=True)
def nb_rvs(U, mu, b, kappa):
    if -kappa < U < 1 / kappa:
        sgn = np.sign(U)
        return mu - (b / (sgn * kappa ** sgn)) * np.log(1 - U * sgn * kappa ** sgn)
    return nb_ppf(U, mu, b, kappa)
```
And call the nb_rvs in `rvs`

tried testing it out as well
```
asm = AsymmetricLaplace(kappa=2, mu=0, b=1)
asm1 = laplace_asymmetric(kappa=2, loc=0, scale=1)
print(asm.rvs(10, random_state=np.random.default_rng(1)))
print(asm1.rvs(10,random_state=np.random.default_rng(1)))

output
[-0.89327111  0.69780578  0.17006291  0.67982083  0.488635    0.93752564
  0.07454764  0.85296965 -0.75086494  0.02834773]
[-0.89327111  0.69780578 -3.42738124  0.67982083 -1.88429781 -1.2729362
  0.07454764 -1.34081961 -0.75086494 -6.73655692]
```

preliz/distributions/asymmetric_laplace.py
rohanbabbar04: Done.

preliz/distributions/asymmetric_laplace.py
rohanbabbar04: Done.

preliz/distributions/asymmetric_laplace.py
aloctavodia: You will need to do something like this, otherwise numba is going to complain about the random uniform or something like that


def rvs(mu, b, kappa, size=1, random_state=None):
    random_state = np.random.default_rng(random_state)
    u = np.random.uniform(-kappa, 1/kappa, size,)
    return nb_rvs(u, mu, b, kappa)

@nb.njit()
def nb_rvs(u, mu, b, kappa):
    s = np.sign(u)
    return mu - (1 / (1/b * s * kappa**s)) * np.log(1 - u * s * kappa**s)
    
    
* Set the seed following the test.
* Notice that you need to sample uniformly in the range [-kappa, 1/kappa] and transform all the samples. not sample U(0, 1) and transform the ones inside [-kappa, 1/kappa]

rohanbabbar04: Hmm...
I did that so that we could test our function with `scipy.stats`, since they just use `uniform(0, 1)` and then apply `ppf`.
Now since we will be changing the range of uniform, should I skip this test for `AsymmetricLaplace`

aloctavodia: I made a mistake in the previous snippet. This should give exactly the same result compared to scipy (up to machine precision)

def rvs(mu, b, kappa, size=1, random_state=None):
    rng = np.random.default_rng(random_state)
    u = rng.uniform(-kappa, 1/kappa, size,)
    return nb_rvs(u, mu, b, kappa)

@nb.njit()
def nb_rvs(u, mu, b, kappa):
    s = np.sign(u)
    return mu - (1 / (1/b * s * kappa**s)) * np.log(1 - u * s * kappa**s)

rohanbabbar04: @aloctavodia we will be using ` u = rng.uniform(-kappa, 1/kappa, size,)`, whereas scipy uses `rng.uniform(size=size)` and then apply ppf.
When we set the range to [-kappa, 1/kappa] we will have negative values as well which will be calculated in our case whereas that is not the case in the `scipy.stats` one.

aloctavodia: But we are not going to use the ppf (that is only defined for the interval [0, 1]). We are going to transform those uniform values using the nb_rvs function.

rohanbabbar04: Yeah I know, I am asking about the `test_scipy.py` tests, there we need to skip the tests for `rvs`

aloctavodia: No need to skip it, I already make the comparison and is essentially the same

rohanbabbar04: Thanks, I realized my mistake...
I realized that we are trying to map our `[-kappa, 1/kappa]` uniform distribution to scipy's `[0,1]`.

I was trying to equate the answer of each U value and that's where I was getting it wrong..

rohanbabbar04:
Okay a test failure occurred

```
FAILED preliz/tests/test_distributions.py::test_mle[ZeroInflatedNegativeBinomial-params40] - AssertionError: 
Arrays are not almost equal to 1 decimals
 ACTUAL: 5.6
 DESIRED: 6.363206728786223
```

PR #391, opened 2024-04-12
Add faster Triangular

rohanbabbar04:
## Description
- Add `Triangular` with all methods.
- Add test in `test_scipy.py`
- Add to docs

## Checklist


preliz/distributions/triangular.py
```diff
@@ -0,0 +1,223 @@
...
+    def entropy(self):
+        return nb_entropy(self.lower, self.upper)
+
+    def mean(self):
+        return (self.lower + self.c + self.upper) / 3
+
+    def median(self):
+        if self.c >= (self.lower + self.upper) / 2:
+            return self.lower + ((self.upper - self.lower) * (self.c - self.lower) / 2) ** 0.5
+        return self.upper - ((self.upper - self.lower) * (self.upper - self.c) / 2) ** 0.5
```
aloctavodia, line 123: Use np.where, so this becomes vectorized. Not sure how useful is to support moments for distributions with parameters being arrays. But we do for many distributions. At some point we should discuss if we want this.

rohanbabbar04: Sure, we can discuss on that
I have changed it to `np.where` as of now

PR #394, opened 2024-04-12
Add faster Logistic

rohanbabbar04:
## Description
- Add `Logistic` with all methods
- Add test in `test_scipy.py`.
- Add Logistic to docs

## Checklist

### DIFF
--- a/preliz/distributions/logistic.py
+++ b/preliz/distributions/logistic.py
@@ -124,9 +124,8 @@ class Logistic(Continuous):
         return random_state.logistic(self.mu, self.s, size)
 
     def _fit_moments(self, mean, sigma):
-        mu = mean
         s = (3 * sigma**2 / np.pi**2) ** 0.5
-        self._update(mu, s)
+        self._update(mean, s)
 
     def _fit_mle(self, sample, **kwargs):
         optimize_ml(self, sample)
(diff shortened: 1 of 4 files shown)


preliz/distributions/logistic.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+@nb.njit(cache=True)
+def nb_ppf(q, mu, s):
+    return mu + s * np.log(q / (1 - q))
+
+
+@nb.njit(cache=True)
+def nb_entropy(s):
+    return np.log(s) + 2
```
aloctavodia, line 157: I think we do not have a strict order, but usually it's cdf, ppf, entropy logpf, neg_logpdf

preliz/distributions/logistic.py
```diff
@@ -0,0 +1,157 @@
...
+        return 6 / 5
+
+    def rvs(self, size=None, random_state=None):
+        random_state = np.random.default_rng(random_state)
+        return random_state.logistic(self.mu, self.s, size)
+
+    def _fit_moments(self, mean, sigma):
+        mu = mean
+        s = (3 * sigma**2 / np.pi**2) ** 0.5
+        self._update(mu, s)
```
aloctavodia, line 129:
```suggestion
    def _fit_moments(self, mean, sigma):
        s = (3 * sigma**2 / np.pi**2) ** 0.5
        self._update(mean, s)
```

PR #401, opened 2024-04-16
Add faster Cauchy

rohanbabbar04:
## Description
- Add `Cauchy` with all methods
- Add test in `test_scipy.py`.
- Add Cauchy to docs

## Checklist

### DIFF
--- a/preliz/distributions/cauchy.py
+++ b/preliz/distributions/cauchy.py
@@ -1,0 +1,162 @@
+# pylint: disable=attribute-defined-outside-init
+# pylint: disable=arguments-differ
+import numpy as np
+import numba as nb
+
+from ..internal.optimization import optimize_ml
+from ..internal.special import ppf_bounds_cont
+from ..internal.distribution_helper import all_not_none, eps
+from .distributions import Continuous
+
+
+class Cauchy(Continuous):
+    r"""
+    Cauchy Distribution
+
+    The pdf of this distribution is
+
+    .. math::
+
+        f(x \mid \alpha, \beta) =
+            \frac{1}{\pi \beta [1 + (\frac{x-\alpha}{\beta})^2]}
+
+    .. plot::
+        :context: close-figs
+
+        import arviz as az
+        from preliz import Cauchy
+        az.style.use('arviz-doc')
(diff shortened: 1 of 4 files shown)


preliz/distributions/cauchy.py
```diff
@@ -0,0 +1,162 @@
...
+    def skewness(self):
+        return np.nan
+
+    def kurtosis(self):
+        return np.nan
+
+    def rvs(self, size=None, random_state=None):
+        random_state = np.random.default_rng(random_state)
+        random_samples = random_state.uniform(0, 1, size)
+        return nb_rvs(random_samples, self.alpha, self.beta)
```
aloctavodia, line 125: The standard cauchy is available from numpy you just need to scale it and shift it https://numpy.org/doc/stable/reference/random/generated/numpy.random.standard_cauchy.html

preliz/distributions/cauchy.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@nb.njit(cache=True)
+def nb_logpdf(x, alpha, beta):
+    return -np.log(np.pi) - np.log(beta) - np.log(1 + ((x - alpha) / beta) ** 2)
+
+
+@nb.njit(cache=True)
+def nb_neg_logpdf(x, mu, s):
+    return -(nb_logpdf(x, mu, s)).sum()
```
aloctavodia, line 157:
```suggestion
@nb.njit(cache=True)
def nb_neg_logpdf(x, alpha, beta):
    return -(nb_logpdf(x, alpha, beta)).sum()
```

preliz/distributions/cauchy.py
rohanbabbar04: For this i did
`self.alpha + self.beta * random_state.standard_cauchy(size)`

but my answer was not matching, can you please correct me here

aloctavodia: this is right. Maybe results do not match at the individual level, i.e. you are not getting exactly the same sequence of samples, the distributions should match. Anyway, tests are going to be simple if you keep this unchanged.

aloctavodia (approved):

PR #414, opened 2024-04-20
Add faster Truncated Normal

rohanbabbar04:
## Description
- Add `TruncatedNormal` with all methods.
- Add test in test_scipy.py.
- Add `TruncatedNormal` to docs.

Don't really know why the `test_mle` in `preliz/tests/test_distributions.py` is failing, changing it to `(-np.inf,np.inf)` seems to work...

## Checklist


preliz/distributions/truncatednormal.py
```diff
@@ -0,0 +1,415 @@
...
+        random_state = np.random.default_rng(random_state)
+        random_samples = random_state.uniform(0, 1, size)
+        return nb_rvs(random_samples, self.mu, self.sigma, self.lower, self.upper)
+
+    def _fit_moments(self, mean, sigma):
+        # Assume gaussian
+        self._update(mean, sigma)
+
+    def _fit_mle(self, sample):
+        optimize_ml(self, sample)
```
aloctavodia, line 352:
```suggestion
    def _fit_mle(self, sample):
        mean, sigma = mean_and_std(sample)
        self._update(mean, sigma, np.min(sample), np.max(sample))
        optimize_ml(self, sample)
```

Where `mean_and_std` is a function from preliz.internal.special

aloctavodia: This should fix the issue, actually setting the mean and sigma is not needed. This should also work with self._update(None, None, np.min(sample), np.max(sample))

preliz/tests/test_distributions.py
```diff
@@ -161,7 +161,7 @@ def test_moments(distribution, params):
         (StudentT, (4, 0, 1)),
         (StudentT, (1000, 0, 1)),
         (Triangular, (-3, 0, 5)),
-        (TruncatedNormal, (0, 1, -1, 1)),
+        (TruncatedNormal, (0, 1, -np.inf, np.inf)),
```
aloctavodia, line 164: revert

rohanbabbar04: Done...

preliz/distributions/truncatednormal.py
rohanbabbar04: Thanks this seems to work...

PR #442, opened 2024-05-18
Adds support for multiple targets with weights to ppe

### LINKED ISSUE #427: PPE: Accept more than one distribution as target
Currently PPE, only accepts as a target a single PreliZ distribution. We should allow more than one distribution. I see at least two useful options

* A list, and then we assume all distributions have an equal contribution. i.e. we sample all distributions with the same probability.
* Or a list of tuples, like [(dist0, w0), (dist1, w1)], for unequal contribution. i.e. we sample all distributions proportional to the weight `w`

To implement this we need to change this line https://github.com/arviz-devs/preliz/blob/main/preliz/internal/optimization.py#L235 into a function that takes `target` and returns a sample or we can precompute the samples that we pass to the optimization routine (the better option is probably the best one).

rohanbabbar04:
Closes #427 

## Description
- Add support for `target` as a list of tuples with each tuple having PreliZ distribution and weight.
- Add parameter `seed` to ppe.
- Add tests to `test_ppe`.
- Update docs.

## Checklist


preliz/internal/optimization.py
```diff
@@ -227,12 +227,15 @@ def optimize_beta_mode(lower, upper, tau_not, mode, dist, mass, prob):
             tau_not += 0.5 * tau_not
 
 
-def optimize_pymc_model(fmodel, target, draws, prior, initial_guess, bounds, var_info, p_model):
+def optimize_pymc_model(fmodel, targets, draws, prior, initial_guess, bounds, var_info, p_model):
```
aloctavodia, line 230: target is ok. as the target can be a mixture of distributuons

preliz/internal/optimization.py
```diff
@@ -227,12 +227,15 @@ def optimize_beta_mode(lower, upper, tau_not, mode, dist, mass, prob):
...
     for _ in range(400):
         # can we sample systematically from these and less random?
         # This should be more flexible and allow other targets than just
         # a preliz distribution
-        obs = target.rvs(draws, random_state=np.random.default_rng(_))
+        if isinstance(targets, list):
+            obs = get_weighted_rvs([d[0] for d in targets], [w[1] for w in targets], draws, _)
```
aloctavodia, line 236: better to pass target to this function and extract the info internally

preliz/internal/optimization.py
```diff
@@ -452,3 +455,9 @@ def func(x, dist, q):
         left, right = right, right * factor
 
     return brentq(func, left, right, args=(dist, q))
+
+
+def get_weighted_rvs(targets, weights, size, seed):
+    target_rnd_choices = np.random.choice(len(targets), size=size, p=weights)
+    samples = [target.rvs(size, np.random.default_rng(seed)) for target in targets]
+    return np.choose(target_rnd_choices, samples)
```
aloctavodia, line 463: create a random generator once and then use it everywhere. I think it should be generated inside ppe.py and then propagated as needed.

preliz/internal/optimization.py
rohanbabbar04: Changed it to `target`

preliz/internal/optimization.py
rohanbabbar04: Done

preliz/internal/optimization.py
rohanbabbar04: For this, I think rng needs to be defined in `optimize_pymc_model`...

aloctavodia: Ok but the user should be able to pass a seed to ppe

rohanbabbar04: Hmm...
This sounds good, we can allow the user to input a seed keeping default to be 0.

PR #450, opened 2024-06-01
Add pymc support to predictive explorer

rohanbabbar04:
## Description
 - Add pymc support to predictive_explorer.
 - Add parameter `engine` to predictive_explorer.
 - Add test in test_predictive_explorer.py.
- Update docs

## Checklist

### DIFF
--- a/preliz/internal/plot_helper.py
+++ b/preliz/internal/plot_helper.py
@@ -469,7 +469,6 @@ def plot_repr(results, kind_plot, references, iterations, ax):
                 ax.set_xticks(bins + 0.5)
         else:
             bins = "auto"
-        print(results.T.shape)
         ax.hist(
             results.T,
             alpha=alpha,
--- a/preliz/internal/plot_helper.py
+++ b/preliz/internal/plot_helper.py
@@ -5,10 +5,10 @@ import sys
 try:
     from IPython import get_ipython
     from ipywidgets import FloatSlider, IntSlider, FloatText, IntText, Checkbox, ToggleButton
+    from pymc import sample_prior_predictive
 except ImportError:
     pass
 
-from pymc import sample_prior_predictive
 from arviz import plot_kde, plot_ecdf, hdi, extract
 from arviz.stats.density_utils import _kde_linear
 import numpy as np
--- a/preliz/internal/plot_helper.py
+++ b/preliz/internal/plot_helper.py
@@ -15,6 +15,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib import _pylab_helpers, get_backend
 from matplotlib.ticker import MaxNLocator
+from .logging import disable_pymc_sampling_logs
 
 
 def plot_pointinterval(distribution, interval="hdi", levels=None, rotated=False, ax=None):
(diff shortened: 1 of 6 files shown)


preliz/internal/plot_helper.py
```diff
@@ -467,6 +501,50 @@ def plot_repr(results, kind_plot, references, iterations, ax):
...
+    alpha = max(0.01, 1 - iterations * 0.009)
+
+    if kind_plot == "hist":
+        if results[0].dtype.kind == "i":
+            bins = np.arange(np.min(results), np.max(results) + 1.5) - 0.5
+            if len(bins) < 30:
+                ax.set_xticks(bins + 0.5)
+        else:
+            bins = "auto"
+        reshaped_t = np.reshape(results.T, (-1, results.shape[0]))
```
aloctavodia, line 514:
```suggestion

```

preliz/internal/plot_helper.py
```diff
@@ -467,6 +501,50 @@ def plot_repr(results, kind_plot, references, iterations, ax):
...
+        ax.hist(
+            reshaped_t,
```
aloctavodia, line 516:
```suggestion
            results,
```

aloctavodia (changes requested):

aloctavodia:
sample_prior_predictive emits a message. saying "Samplig", plus a list of the sampled variables. We should disable that message. It is generated with the logging module. We should suppress it locally with a context manager, or disable it before sampling and enable it again after sampling. We need to do this to avoid messing up calls to `sample_prior_predictive` outside of predictive_explorer

aloctavodia:
It is no needed for this PR, but to think about it we may want to try to guess the value of "engine", for instance for a pymc model we could inspect the source and detect the regex pattern for "with pm.Model() as" where the particle "pm." is something that can vary.

We may want to keep the "engine" argument and default to "auto" or something similar, just in case our automatic detection fails

preliz/internal/plot_helper.py
```diff
@@ -425,6 +426,39 @@ def looper(*args, **kwargs):
...
+            auto = False
+
+        for _ in range(iterations):
+            with func(*args, **kwargs) as model:
+                obs_name = model.observed_RVs[0].name
+                idata = sample_prior_predictive()
+                val = extract(idata, group="prior_predictive")[obs_name].values
+            if not np.isnan(val).any():
+                results.append(val)
+        results = np.array(results)
```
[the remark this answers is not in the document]
rohanbabbar04: Got it 
Thanks...

rohanbabbar04:
> sample_prior_predictive emits a message. saying "Samplig", plus a list of the sampled variables. We should disable that message. It is generated with the logging module. We should suppress it locally with a context manager, or disable it before sampling and enable it again after sampling. We need to do this to avoid messing up calls to `sample_prior_predictive` outside of predictive_explorer

Created a contextmanager to handle the logs generated during this step.

rohanbabbar04:
> It is not needed for this PR, but something to think about.
> 
> We may want to try to guess the value of "engine", for instance for a pymc model we could inspect the source and detect the regex pattern for "with pm.Model() as" where the particle "pm." is something that can vary. For Bambi it should also be simple to detect it.
> 
> We may want to keep the "engine" argument and default to "auto" or something similar, just in case our automatic detection fails

Sounds good.... Let me create a new issue for this

rohanbabbar04:
I am currently keeping the sampling pymc code separated from the preliz code, like it was discussed in the jupyter notebook.
Better to not disrupt the current implementation of the already existing preliz code.

aloctavodia:
agreed, we can refactor and combine later

rohanbabbar04:
> agreed, we can refactor and combine later. Something really easy to combine is the plotting part, right?

Yes...

PR #455, opened 2024-06-06
Add support for engine="auto" and bambi models in predictive explorer

### LINKED ISSUE #451: Update engine in `predictive_explorer` to handle auto detection of fmodel
A enhancement after merging #450 
## Tell us about it

 - Auto detected the type of model inputted by the user.
 - We keep `engine=auto` for auto detection.

### LINKED ISSUE #453: Add support to Bambi models in Predictive Explorer
We already have PreliZ and PyMC, we should have Bambi too.

rohanbabbar04:
Closes #453 
Closes #451 
## Description
- Add support when `engine="auto"`
- Add support for Bambi Models.(Still need to write a case when the Priors are not provided by the user)
- Add docs and test.
- I changed `extract(idata, group="prior_predictive")[model.response_name].values` to `extract(idata, group="prior_predictive")[model.response_name].values.T`. This I think I missed in pymc but I have corrected it in this PR.

## Checklist

### DIFF
--- a/preliz/internal/parser.py
+++ b/preliz/internal/parser.py
@@ -4,8 +4,6 @@ import re
 from sys import modules
 
 import numpy as np
-import pymc as pm
-import bambi as bmb
 
 from preliz import distributions
 from .distribution_helper import init_vals
@@ -20,9 +18,9 @@ def inspect_source(fmodel):
         for name, param in signature.parameters.items()
     }
     model = fmodel(**default_params)
-    if isinstance(model, pm.Model):
+    if getattr(model, "basic_RVs", False):
         engine = "pymc"
-    elif isinstance(model, bmb.Model):
+    elif getattr(model, "formula", False):
         engine = "bambi"
     else:
         engine = "preliz"
--- a/preliz/internal/parser.py
+++ b/preliz/internal/parser.py
@@ -24,11 +24,17 @@ def parse_function_for_pred_textboxes(source, signature, engine="preliz"):
     regex = r"\b" + r"\b|\b".join(slidify) + r"\b"
 
     all_dist_str = dist_as_str()
-    matches = match_preliz_dist(all_dist_str, source, "preliz")
+    matches = match_preliz_dist(all_dist_str, source, engine)
 
     for match in matches:
-        dist_name_str = match.group(2)
-        arguments = [s.strip() for s in match.group(3).split(",")]
+        if engine == "bambi":
+            dist_name_str = match.group(1)
+        else:
+            dist_name_str = match.group(2)
+        if engine == "bambi":
+            arguments = [s.strip() for s in match.group(2).split(",")]
+        else:
+            arguments = [s.strip() for s in match.group(3).split(",")]
         if engine == "pymc":
             args = pymc_parse_arguments(arguments, regex)
         else:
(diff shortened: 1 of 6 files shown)


preliz/internal/predictive_helper.py
```diff
@@ -68,3 +69,12 @@ def select_prior_samples(selected, prior_samples, model):
     subsample = {rv: prior_samples[rv][selected] for rv in model.keys()}
 
     return subsample
+
+
+def get_engine(source):
+    source = re.sub(r"#.*$|^#.*$", "", source, flags=re.MULTILINE)
+    if re.search(r"\s*(?:bmb|bambi)\.Model\s*", source):
+        return "bambi"
+    if re.search(r"\s*(?:pm|pymc)\.Model\s*", source):
```
aloctavodia, line 78: This will fail if the user non-standard imports, which may be more common than we are expecting. I think we need a slightly more fancy approach. Something like this 

def get_returned_object_class(func):
    sig = inspect.signature(func)
    default_params = {
        k: (v.default if v.default is not inspect.Parameter.empty else np.nan)
        for k, v in sig.parameters.items()
    }
    return func(**default_params).__class__
    
Maybe the logic of this function should be part of `inspect_source` and should also have the logic to directly return the engine. That is the new `inspect_source`, should return 3 objects source, signature, engine

preliz/predictive/predictive_explorer.py
```diff
@@ -20,8 +26,8 @@ def predictive_explorer(
     Parameters
     ----------
     fmodel : callable
-        A function with PreliZ distributions or PyMC distributions, depending on the selected
-        engine. The PreliZ distributions should call their rvs method.
+        A function with PreliZ distributions, PyMC distributions, or Bambi distributions.
+        The PreliZ distributions should call their rvs method.
```
aloctavodia, line 30:
```suggestion
        A function with a PreliZ model, a PyMC model, or a Bambi model. See examples section below for details
```

aloctavodia: This could be done in a separated PR. We should add an example section to the docstring. The section should describe a minimal example using preliz, another using pymc and another bambi.

aloctavodia (changes requested):

preliz/internal/plot_helper.py
```diff
@@ -427,9 +427,36 @@ def looper(*args, **kwargs):
...
+            x_min = None
+            x_max = None
+            auto = True
+        else:
+            auto = False
+
+        model = func(*args, **kwargs)
+        model.build()
+        with disable_pymc_sampling_logs():
+            idata = model.prior_predictive(iterations, random_seed=iterations)
```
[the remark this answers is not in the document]
rohanbabbar04: Agreed, i think we can remove the random_seed parameter

preliz/internal/parser.py
```diff
@@ -4,6 +4,8 @@
 from sys import modules
 
 import numpy as np
+import pymc as pm
+import bambi as bmb
```
aloctavodia, line 8: These 2 imports should be optional

preliz/internal/parser.py
```diff
@@ -13,8 +15,18 @@ def inspect_source(fmodel):
...
-
-    return source, signature
+    default_params = {
+        name: (param.default if param.default is not inspect.Parameter.empty else np.nan)
+        for name, param in signature.parameters.items()
+    }
+    model = fmodel(**default_params)
+    if isinstance(model, pm.Model):
+        engine = "pymc"
+    elif isinstance(model, bmb.Model):
```
aloctavodia, line 25: This is the right thing to do in general, but this will require the user to have pymc and bambi to be installed.

PR #465, opened 2024-06-12
Add Distribution Gallery 

aleicazatti:
## Description

* Add Distribution Gallery grid
* Add Normal distribution page as example

Notes: 

I didn’t know what to add as a Bayesian story or application initially. Then I thought we could mention something in the context of its use as priors. Feel free to propose changes or mention other Bayesian-focused stories, applications, explanations, or use cases.


aloctavodia:
you need to add `sphinx_design` to `requirements-docs.txt`

docs/examples/normal_distribution.rst
```diff
@@ -0,0 +1,59 @@
...
+____________________________________________________
+
+========  ==========================================
+Support   :math:`x \in \mathbb{R}`
+Mean      :math:`\mu`
+Variance  :math:`\sigma^2`
+========  ==========================================
+
+Relevant mathematical formulas:
+____________________________________________________
```
aloctavodia, line 32:
```suggestion
```

docs/examples/normal_distribution.rst
```diff
@@ -0,0 +1,59 @@
...
+
+The normal distribution is often the result of summing many small, independent effects, a concept encapsulated by the Central Limit Theorem. Many biological processes exhibit this property, such as human height, blood pressure, and measurement errors. In such cases, numerous genetic, environmental, and random factors contribute to the observed outcome, each adding a small effect. For example, human height is influenced by multiple genes and environmental factors, which together result in a normal distribution of heights across a population.
+
+Normal priors are commonly chosen in Bayesian analysis because they represent a state of limited prior knowledge or weak information. When we assume a parameter has finite variance but lack strong prior knowledge, the normal distribution becomes a suitable choice. This is due to the normal distribution's maximum entropy property among all distributions with a specified mean and variance, ensuring it introduces the least amount of additional assumptions. 
+
+Probability Density Function (PDF):
+____________________________________________________
+
+.. image:: normal_pdf.png
```
aloctavodia, line 13: We should include the code to generate the images (instead of the images), or we should include somewhere the instructions to generate new images, like using "arviz-docs" style or whatever..

docs/examples/normal_distribution.rst
```diff
@@ -0,0 +1,59 @@
...
+    \exp\left\{ -\frac{1}{2} \left(\frac{x-\mu}{\sigma}\right)^2 \right\}
+
+**Cumulative Distribution Function (CDF):**
+
+.. math::
+
+    F(x \mid \mu, \sigma) =
+    \frac{1}{2} \left[ 1 + \text{erf} \left( \frac{x - \mu}{\sigma \sqrt{2}} \right) \right]
+
+where erf is the error function.
```
aloctavodia, line 49: maybe a link to the wikipedia page for the erf function

docs/conf.py
```diff
@@ -51,6 +51,8 @@
     "myst_nb",
     "matplotlib.sphinxext.plot_directive",
     "sphinx_tabs.tabs",
+    "pydata_sphinx_theme",
```
OriolAbril, line 54: this is already in the requirements so it is installed and is the value of `html_theme`. It should not be added as an extension. The best case scenario is nothing happens, worst case scenario is it breaks the doc build

OriolAbril (commented):
regarding the image generation, as I assume it won't take long to do and preliz is already installed in the doc build, we could have the pages be markdown notebooks that include the code and are executed when building the docs. The code can be removed from the rendered page or hidden inside a dropdown.

I can convert one of the pages into a markdown notebook to show how it would look like.

aleicazatti:
I applied the suggested changes, except the code to generate the images so we do what @OriolAbril suggests.

aloctavodia:
I see math rendering issues
Captura desde 2024-06-13 13-20-31

OriolAbril:
> I see math rendering issues

I forgot to activate dollarmath parsing, should be fixed now

docs/examples/normal_distribution.md
```diff
@@ -0,0 +1,88 @@
...
+
+The normal distribution is often the result of summing many small, independent effects, a concept encapsulated by the Central Limit Theorem. Many biological processes exhibit this property, such as human height, blood pressure, and measurement errors. In such cases, numerous genetic, environmental, and random factors contribute to the observed outcome, each adding a small effect. For example, human height is influenced by multiple genes and environmental factors, which together result in a normal distribution of heights across a population.
+
+Normal priors are commonly chosen in Bayesian analysis because they represent a state of limited prior knowledge or weak information. When we assume a parameter has finite variance but lack strong prior knowledge, the normal distribution becomes a suitable choice. This is due to the normal distribution's maximum entropy property among all distributions with a specified mean and variance, ensuring it introduces the least amount of additional assumptions. 
+
+## Probability Density Function (PDF):
+
+```{code-cell}
+---
+tags: [hide-input]
```
OriolAbril, line 23: I used hide here and remove below so we can compare. For this type of doc remove might make more sense

OriolAbril:
The main limitation I am thinking about now are the images for the grid view. Maybe the pdf one could be saved as part of the code in the notebook but not commited to the repo (`docs/examples/*.png` in gitignore would take care of that for example). Not sure in what order do things happen though, if the notebooks are executed as the docs are built instead of before it might not work.

Another option would be to save the grid images independently and if needed (manually run the notebooks to save them again and update them).

docs/examples/normal_distribution.md
aloctavodia: Yeah, remove is better for this one

docs/gallery_content.rst
```diff
@@ -0,0 +1,622 @@
...
+      :shadow: none
+      :class-card: example-gallery
+
+      .. image:: examples/placeholder.png
+         :alt: Multivariate Normal
+
+      +++
+      Multivariate Normal
+
+Special Distributions
```
aloctavodia, line 583: Distribution Modifiers

aloctavodia:
Is this still WIP or ready to merge?

aleicazatti:
Sorry for the delay, if there aren't any other suggestions I think is ready to merge.

Issue #475, opened 2024-06-19
Distributions Gallery: Normal


aloctavodia:
Add these distributions to "see also". 

* TruncatedNormal
* VoMises (because is like a normal but on the circle)
* Logistic (because it "looks like a normal" and is sometimes used as a robust alternative)

What about these two? They are related to the normal, but not that close in the "prior-space" 

* HalfNormal
* LogNormal

OriolAbril:
We could have different sections/boxes instead of only a "see also". For example: "common alternatives", where we list a distribution with a link to the corresponding preliz page and then a single line with the main difference, and "related distributions"/keep the see also, with a more flexible format.

PR #477, opened 2024-06-19
Update normal distribution page

aleicazatti:
## Description

Update normal distribution page based on suggestions from #475

### DIFF
--- a/docs/examples/normal_distribution.md
+++ b/docs/examples/normal_distribution.md
@@ -81,9 +81,21 @@ $$
 where [erf is the error function](https://en.wikipedia.org/wiki/Error_function).
 
 
-:::{seealso}
-{doc}`students_t_distribution`
-:::
+```{seealso}
+:class: seealso
+
+**Common Alternatives:**
+
+- [Truncated Normal Distribution](truncated_normal_distribution.md) - Limits the normal distribution to values within a specified range.
+- [Half Normal Distribution](half_normal_distribution.md) - Considers only the positive half of the normal distribution.
+- [Log Normal Distribution](log_normal_distribution.md) - Assumes the logarithm of values follows a normal distribution.
+- [Student's T Distribution](students_t_distribution.md) - Used instead of the normal distribution when the sample size is small.
+
+**Related Distributions:**
+
+- [Von Mises Distribution](vonmises_distribution.md) - Suitable for data on a circular scale, like angles or time of day.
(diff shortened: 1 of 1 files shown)


docs/examples/normal_distribution.md
```diff
@@ -81,9 +81,21 @@ $$
...
+
+- [Truncated Normal Distribution](truncated_normal_distribution.md) - Limits the normal distribution to values within a specified range.
+- [Half Normal Distribution](half_normal_distribution.md) - Considers only the positive half of the normal distribution.
+- [Log Normal Distribution](log_normal_distribution.md) - Assumes the logarithm of values follows a normal distribution.
+- [Student's T Distribution](students_t_distribution.md) - Used instead of the normal distribution when the sample size is small.
+
+**Related Distributions:**
+
+- [Von Mises Distribution](vonmises_distribution.md) - Suitable for data on a circular scale, like angles or time of day.
+- [Logistic Distribution](logistic_distribution.md) - Has heavier tails than the normal distribution, used in logistic regression.
```
aloctavodia, line 97:
```suggestion
- [Student's T Distribution](students_t_distribution.md) - Commonly used as a robust alternative to the Normal, it has a third parameter controlling the heaviness of the tails.
- [Logistic Distribution](logistic_distribution.md) - Used as a robust alternative to the Normal, it has heavier tails than the normal distribution.
- [Truncated Normal Distribution](truncated_normal_distribution.md) - Bounds the Normal distribution within a specified range. Sometimes, it is used for Normal-like parameters when certain values are not permitted.

**Related Distributions:**

- [Half Normal Distribution](half_normal_distribution.md) - Considers only the positive half of the Normal distribution.
- [Log Normal Distribution](log_normal_distribution.md) - If a random variable follows a LogNormal distribution, its natural logarithm will be Normally distributed. It is suitable for modelling positive and right-skewed data or parameters.
- [Von Mises Distribution](vonmises_distribution.md) - Close to a wrapped Normal on the circle. Suitable for data on a circular scale, like angles or time of day.

```

docs/examples/normal_distribution.md
```diff
@@ -81,9 +81,21 @@ $$
 where [erf is the error function](https://en.wikipedia.org/wiki/Error_function).
```
aloctavodia, line 81:
```suggestion
where [erf](https://en.wikipedia.org/wiki/Error_function) is the error function.
```

PR #495, opened 2024-07-11
don't run tests if only docs folder changes

OriolAbril:
I think this will prevent tests being executed when there are only changes in `docs/`


aloctavodia:
thanks!

PR #505, opened 2024-07-21
Add tabs for predictive_explorer in observed_space_examples_all.rst

### LINKED ISSUE #473: Add 3 tabs for predictive_explorer example
It would be nice to have 3 tabs, one for a "preliz model", one for a PyMC model and one for the Bambi model, when rendering the docs I think it is possible to run only one tab

rohanbabbar04:
Closes #473 
## Description
- Added `observed_space_examples_all.rst` which includes the tab feature.
- Add images which are used in the example

The issue while using {code-cell} to run the example is that code-cell is not supported when used inside another item(in our case tabs)...The best possible solution is to use code-block and add images

## Checklist


docs/index.rst
```diff
@@ -121,7 +121,7 @@ Contents
    :maxdepth: 2
 
    examples/param_space_1d_examples
-   examples/observed_space_examples
+   examples/observed_space_examples_all
```
aloctavodia, line 124: Why the rename?

PR #507, opened 2024-08-02
Distributions Gallery: Add ExGaussian

aleicazatti:
## Notes:

I tried to derive the PDF and CDF from the equations on the [Wikipedia page](https://en.wikipedia.org/wiki/Exponentially_modified_Gaussian_distribution) using $\nu$, but in  the docstring of the distribution, the PDF expression is a little different.

### DIFF
--- a/docs/examples/gallery/exgaussian.md
+++ b/docs/examples/gallery/exgaussian.md
@@ -65,18 +65,17 @@ f(x \mid \mu, \sigma, \nu) =
 \frac{1}{\nu}\exp\left(\frac{\mu-x}{\nu}+\frac{\sigma^2}{2\nu^2}\right)\Phi\left(\frac{x-\mu}{\sigma}-\frac{\sigma}{\nu}\right)
 $$
 
-where $\Phi$ is the [standard normal PDF](normal.md).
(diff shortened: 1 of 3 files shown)


aloctavodia:
Are the differences indicate a potential mistake somewhere or just two ways to express the same? Did you compare the mathematical expressions for the pdf and cdf with the code for them? Are

aleicazatti:
I was checking. It seems they are equivalent. I dont' know if we should use the docstring/code expressions or the ones in Wikipedia.

aloctavodia:
Not sure, probably match the code.

docs/examples/gallery/exgaussian.md
```diff
@@ -0,0 +1,92 @@
...
+---
+tags: [remove-input]
+mystnb:
+  image:
+    alt: Ex-Gaussian Distribution PDF
+---
+
+import arviz as az
+from preliz import ExGaussian
+az.style.use('arviz-doc')
```
aloctavodia, line 29:
```suggestion
from preliz import ExGaussian, style
style.use('preliz-doc')
```

PR #508, opened 2024-08-03
Add backfitting with posterior sampling

rohanbabbar04:
## Description
Initial sample for parsing the model and applying posterior backfitting
- Apply code for backfitting.
- Using ancestors in `non_constant_parents` to calculate parent inputs

## Checklist

### DIFF
--- a/preliz/ppls/pymc_io.py
+++ b/preliz/ppls/pymc_io.py
@@ -217,7 +217,7 @@ def parse_backfitting(model, data):
 
     model_info = get_model_information(model)[2]
     parsed_info = [(dist, var) for var, dist in model_info.items()]
-    print(posterior, parsed_info)
+
     new_priors = []
     for dist, var in parsed_info:
         dist._fit_mle(posterior[var].values)
--- a/preliz/ppls/pymc_io.py
+++ b/preliz/ppls/pymc_io.py
@@ -3,7 +3,6 @@
 # pylint: disable=protected-access
 from sys import modules
 
-import preliz as pz
 import numpy as np
 
 try:
@@ -15,6 +14,8 @@ except ModuleNotFoundError:
     pass
 
 from preliz.internal.optimization import get_distributions
+from preliz.distributions import Gamma, Normal, HalfNormal
+from preliz.unidimensional.mle import mle
 
 
 def backfitting(prior, p_model, var_info2):
@@ -236,13 +237,13 @@ def posterior_to_prior(model, posterior, alternative=None):
         dists = [model_info[var]]
 
         if alternative == "auto":
-            dists += [pz.Normal(), pz.HalfNormal(), pz.Gamma()]
+            dists += [Normal(), HalfNormal(), Gamma()]
         elif isinstance(alternative, list):
             dists += alternative
         elif isinstance(alternative, dict):
             dists += alternative.get(var, [])
         # Take the dist with the least penalization term
-        idx = pz.mle(dists, posterior[var].values)[0]
+        idx = mle(dists, posterior[var].values)[0]
         new_priors.append((dists[idx[0]], var))
 
     new_model = "\n".join(f"{var} = {new_prior}" for new_prior, var in new_priors)
--- a/preliz/ppls/pymc_io.py
+++ b/preliz/ppls/pymc_io.py
@@ -4,18 +4,17 @@
 from sys import modules
 
 import preliz as pz
-import arviz as az
 import numpy as np
 
 try:
     from pytensor.tensor import vector, TensorConstant
     from pytensor.graph.basic import ancestors
-    from pymc import logp, compile_pymc, sample
+    from pymc import logp, compile_pymc
     from pymc.util import is_transformed_name, get_untransformed_name
 except ModuleNotFoundError:
     pass
 
-from preliz.internal.optimization import get_distributions, fit_to_sample
+from preliz.internal.optimization import get_distributions
 
 
 def backfitting(prior, p_model, var_info2):
--- a/preliz/ppls/pymc_io.py
+++ b/preliz/ppls/pymc_io.py
@@ -8,6 +8,7 @@ import numpy as np
 
 try:
     from pytensor.tensor import vector, TensorConstant
+    from pytensor.graph.basic import ancestors
     from pymc import logp, compile_pymc
     from pymc.util import is_transformed_name, get_untransformed_name
 except ModuleNotFoundError:
@@ -119,14 +120,10 @@ def get_model_information(model):  # pylint: disable=too-many-locals
     rvs_to_values = model.rvs_to_values
 
     for r_v in model.free_RVs:
-        if not non_constant_parents(r_v):
-            free_rvs.append(r_v)
-
-    for r_v in model.free_RVs:
         r_v_eval = r_v.eval()
         size = r_v_eval.size
         shape = r_v_eval.shape
-        nc_parents = non_constant_parents(r_v)
+        nc_parents = non_constant_parents(r_v, model.free_RVs)
 
         name = r_v.owner.op.name
         dist = pymc_to_preliz[name]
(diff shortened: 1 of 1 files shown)


preliz/ppls/pymc_io.py
```diff
@@ -210,12 +211,40 @@ def non_constant_parents(var_, free_rvs):
...
+    We need to "backfit" because we can not use arbitrary samples as priors.
 
-    model_info = get_model_information(model)[2]
-    parsed_info = [(dist, var) for var, dist in model_info.items()]
+    Parameters
+    ----------
+    model : A PyMC model
+    A probabilistic model
+
+    data : np.ndarray
```
aloctavodia, line 224: The input should be an inferencedata with the posterior group

preliz/ppls/pymc_io.py
```diff
@@ -210,12 +211,40 @@ def non_constant_parents(var_, free_rvs):
...
+    Samples which needs to be fitted using "backfitting"
+
+    alternative : list, defaults to None
```
aloctavodia, line 227: probably a very flexible solution will be to make this accept None, "auto", list or dict

* If None only evaluate original family
* If "auto" evaluate original + a set of default distributions
* If list evaluate original + the distributions inside the list. The list should be valid preliz distributions
* If dict evaluate original + the values in the dict. The keys should be var_names in the model and the values valid preliz distributions.

preliz/ppls/pymc_io.py
```diff
@@ -210,12 +211,40 @@ def non_constant_parents(var_, free_rvs):
...
+    with model:
+        idata = sample(random_seed=4124)
+    posterior = az.extract(idata, group="posterior")
+    model_info = get_model_information(model)[2]
+    if alternative:
+        parsed_info = []
+        for key in model_info.keys():
+            if key in alternative:
+                dists_extra = [pz.Gamma(), pz.Normal(), pz.HalfNormal(), model_info[key]]
+                idx = np.argmin(fit_to_sample(dists_extra, data, data.min(), data.max()).losses)
```
aloctavodia, line 242: Better to use pz.mle function that takes care of this

preliz/ppls/pymc_io.py
```diff
@@ -210,12 +211,40 @@ def non_constant_parents(var_, free_rvs):
...
+    Samples which needs to be fitted using "backfitting"
+
+    alternative : list, defaults to None
+    Users can add the model variables to consider alternative distributions while fitting samples
 
+    """
+
+    data = np.asarray(data)
+    with model:
+        idata = sample(random_seed=4124)
```
aloctavodia, line 234: don't sample, the user should provide the idata with the posterior group

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +200,54 @@ def reshape_params(model, var_info, p_model, params):
...
         if not isinstance(variable, TensorConstant):
-            parents.append(variable.owner.inputs[0])
-
+            for free_rv in free_rvs:
+                if free_rv in list(ancestors([variable])) and free_rv not in parents:
+                    parents.append(free_rv)
     return parents
+
+
+def parse_backfitting(model, data, alternative=None):
```
aloctavodia, line 214: We should have a better name for this. Like "posterior_to_prior" or something like that

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +200,54 @@ def reshape_params(model, var_info, p_model, params):
...
+
+    data : np.ndarray
+    Samples which needs to be fitted using "backfitting"
+
+    alternative : list, defaults to None
+    Users can add the model variables to consider alternative distributions while fitting samples
+
+    """
+
+    data = np.asarray(data)
```
aloctavodia, line 232: what is this data?, the only "data" should be posterior samples

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +200,54 @@ def reshape_params(model, var_info, p_model, params):
...
+    with model:
+        idata = sample(random_seed=4124)
+    posterior = az.extract(idata, group="posterior")
+    model_info = get_model_information(model)[2]
+    if alternative:
```
aloctavodia, line 237: Not sure I fully understand the logic inside the alternative branch. It should be pretty similar to the regular branch except that you should use `pz.mle` instead of  `dist._fit_mle`. Besides with the current logic you are fitting the distributions twice, once inside the alternative branch and then again when you evaluate the distributions inside  `parsed_info`.

preliz/ppls/pymc_io.py
rohanbabbar04: Yes changed it to `posterior` as a parameter

preliz/ppls/pymc_io.py
rohanbabbar04: Have made the changes for the 4 options,
For the 2,3 and 4th we take the term with the least penalization term for each variable

preliz/ppls/pymc_io.py
rohanbabbar04: Yes, pz.mle seems to be the best option

preliz/ppls/pymc_io.py
rohanbabbar04: Have removed `dist._fit_mle` and replaced it with `pz.mle`

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +199,51 @@ def reshape_params(model, var_info, p_model, params):
...
+    for dist, var in parsed_info:
+        dists = [model_info[var]]
+
+        if alternative == "auto":
+            dists += [pz.Normal(), pz.HalfNormal(), pz.Gamma()]
+        elif isinstance(alternative, list):
+            dists += alternative
+        elif isinstance(alternative, dict):
+            dists += alternative.get(var, [])
+        # Take the dist with the least penalization term
```
aloctavodia, line 244: This comment is not accurate

rohanbabbar04: Any better way to write this?

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +199,51 @@ def reshape_params(model, var_info, p_model, params):
...
+        idx = pz.mle(dists, posterior[var].values)[0]
```
aloctavodia: Pass plot=False. Also while this work for a single distribution, in that case it is better to just use the private method fit_mle, it should be slightly faster.

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +199,51 @@ def reshape_params(model, var_info, p_model, params):
...
-
+            for free_rv in free_rvs:
+                if free_rv in list(ancestors([variable])) and free_rv not in parents:
+                    parents.append(free_rv)
     return parents
+
+
+def posterior_to_prior(model, posterior, alternative=None):
+    """
+    Get the model information for fitting the samples from prior into user provided model's prior.
```
aloctavodia, line 216: Fix docstring to reflect what this function is actually doing

preliz/ppls/pymc_io.py
rohanbabbar04: Are you talking about the case when alternative="None", that we should use fit_dist and for other cases use pz.mle(when more than 1 distributions are there)
Since, pz.mle gives us an array with the least loss and we pick the least one?

aloctavodia: Yes, when we only have a single distribution "alternative=None" there is not reason to use use pz.mle other that making the code slightly simpler

preliz/ppls/pymc_io.py
rohanbabbar04: Have updated the docstring

preliz/ppls/pymc_io.py
rohanbabbar04: Thanks have made the changes when `len(dists) = 1`

preliz/ppls/pymc_io.py
```diff
@@ -201,11 +200,55 @@ def reshape_params(model, var_info, p_model, params):
...
+    for dist, var in parsed_info:
+        dists = [model_info[var]]
+
+        if alternative == "auto":
+            dists += [Normal(), HalfNormal(), Gamma()]
+        elif isinstance(alternative, list):
+            dists += alternative
+        elif isinstance(alternative, dict):
+            dists += alternative.get(var, [])
+        if len(dists) == 1:
```
aloctavodia, line 246: Be explicit, one branch if alternative is None and another otherwise. inside the second branch evaluate the 3 other possible values for alternative.

aloctavodia:
I merged because I am starting to be very annoying with the requests, haha. I will submit a PR with the changes I have in mind. You already did a great job.

PR #510, opened 2024-08-09
refactor posterior_to_prior

aloctavodia:
follow up of #508


rohanbabbar04:
This looks good...
Do we include this method in any of the `predictive`?
Also, if this is merged we can start working on the bambi model

aloctavodia:
move it posterior_to_prior to a separate file as this function should be a general interface to more ppls, not just pymc

PR #515, opened 2024-08-13
Update tests for posterior_to_prior

### LINKED ISSUE #513: improve posterior_to_prior test
A very basic test was introduced in #512, we should improve it. For instance, we should check that `pz.posterior_to_prior(model, idata, alternative={"b": [pz.Gamma(mu=0)]})` returns a Gamma distribution for parameter "b" with the mu, sigma parametrization while `pz.posterior_to_prior(model, idata, alternative="auto")` returns the same but with the alpha, and beta parametrization.

rohanbabbar04:
Closes #513 
- Add check for `mu` and `sigma` for `alternative={"b": [pz.Gamma(mu=0)]}`.
- Add check for `alpha` and `beta` for `alternative="auto"`
## Checklist

### DIFF
--- a/preliz/tests/test_posterior_to_prior.py
+++ b/preliz/tests/test_posterior_to_prior.py
@@ -14,10 +14,8 @@ with pm.Model() as model:
 
 def test_p2p():
     pz.posterior_to_prior(model, idata)
-    pattern = r'\s*b = pm.\x1b\[1mGamma\x1b\[0m\("b",\s*alpha\s*=\s*([+-]?\d*\.?\d+),\s*beta\s*=\s*([+-]?\d*\.?\d+)\)'
-    assert bool(re.search(pattern, pz.posterior_to_prior(model, idata, alternative="auto")))
+    assert 'Gamma\x1b[0m("b", alpha=' in pz.posterior_to_prior(model, idata, alternative="auto")
     pz.posterior_to_prior(model, idata, alternative=[pz.LogNormal()])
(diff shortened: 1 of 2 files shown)


preliz/tests/test_posterior_to_prior.py
```diff
@@ -12,6 +14,10 @@
 
 def test_p2p():
     pz.posterior_to_prior(model, idata)
-    pz.posterior_to_prior(model, idata, alternative="auto")
+    pattern = r'\s*b = pm.\x1b\[1mGamma\x1b\[0m\("b",\s*alpha\s*=\s*([+-]?\d*\.?\d+),\s*beta\s*=\s*([+-]?\d*\.?\d+)\)'
+    assert bool(re.search(pattern, pz.posterior_to_prior(model, idata, alternative="auto")))
     pz.posterior_to_prior(model, idata, alternative=[pz.LogNormal()])
-    pz.posterior_to_prior(model, idata, alternative={"b": [pz.Gamma(mu=0)]})
+    pattern = r'\s*b = pm.\x1b\[1mGamma\x1b\[0m\("b",\s*mu\s*=\s*([+-]?\d*\.?\d+),\s*sigma\s*=\s*([+-]?\d*\.?\d+)\)'
```
aloctavodia, line 20: Not against your proposal, but what if test something simpler like`'Gamma\x1b[0m("b", alpha=' in ...` vs `'Gamma\x1b[0m("b", mu=' in ...`

rohanbabbar04: It sounds good, I have made the change

PR #522, opened 2024-08-14
Add bambi support for posterior_to_prior

rohanbabbar04:
## Description
- Add parameter `engine`.
- Add bambi support for `posterior_to_prior`.
- Add test for bambi

## Checklist

### DIFF
--- a/preliz/ppls/bambi_io.py
+++ b/preliz/ppls/bambi_io.py
@@ -1,0 +1,27 @@
+from preliz.ppls.pymc_io import get_model_information
+
+
+def get_bmb_model_information(model):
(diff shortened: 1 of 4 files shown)


preliz/ppls/bambi_io.py
```diff
@@ -0,0 +1,27 @@
+from preliz.ppls.pymc_io import get_model_information
+
+
+def get_bmb_model_information(model):
+    model.build()
```
aloctavodia, line 5: probably better to check model.built

rohanbabbar04: Done :)

PR #566, opened 2024-10-22
Distributions Gallery: Add SkewNormal

aleicazatti:
The PDF is expressed as in the docstring, and the CDF is based on the expression used in the code.

### DIFF
--- a/docs/examples/gallery/skewnormal.md
+++ b/docs/examples/gallery/skewnormal.md
@@ -1,0 +1,127 @@
+---
+jupytext:
+  text_representation:
+    extension: .md
+    format_name: myst
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+# Skew-Normal Distribution
+
+<audio controls> <source src="../../_static/skewnormal.mp3" type="audio/mpeg"> This browser cannot play the pronunciation audio file for this distribution. </audio>
+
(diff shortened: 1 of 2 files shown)


docs/examples/gallery/skewnormal.md
```diff
@@ -0,0 +1,127 @@
...
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+# Skew-Normal Distribution
+
+<audio controls> <source src="../../_static/skewnormal.mp3" type="audio/mpeg"> This browser cannot play the pronunciation audio file for this distribution. </audio>
+
+The skew-normal distribution is a continuous probability distribution that generalizes the normal distribution by introducing a skewness parameter ($\alpha$). This parameter allows the distribution to be asymmetric around its mean, with the skewness determining the direction and degree of the asymmetry. When $\alpha = 0$, the skew-normal distribution reduces to the normal distribution.
```
aloctavodia, line 15:
```suggestion
The skew-normal distribution is a continuous probability distribution that generalizes the normal distribution by introducing a skewness parameter ($\alpha$). This parameter allows the distribution to be asymmetric around its mean, with $\alpha$ determining the direction and degree of the asymmetry. When $\alpha = 0$, the skew-normal distribution reduces to the normal distribution.
```

PR #608, opened 2024-12-17
Distributions Gallery: Add Dirichlet

aleicazatti:


aloctavodia:
The pdf plot emits a few warnings "Ignoring fixed x limits to fulfill fixed data aspect with adjustable data limits". Probably better to catch the warnings so they displayed. Other than that LGTM.

aloctavodia:
I realized the message is not from the warnings module but logging, see https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/axes/_base.py#L2053-L2054. I will try to disable.

PR #625, opened 2025-01-11
Add ruff for linting instead of pylint

### LINKED ISSUE #543: Use Ruff
We should switch to Ruff for linting and code formatting.

We could follow ArviZ:

* https://github.com/arviz-devs/arviz-stats/blob/main/pyproject.toml
* https://github.com/arviz-devs/arviz-stats/blob/main/.pre-commit-config.yaml

rohanbabbar04:
## Description
Closes #543 
- Updated `pyproject.toml` and `.pre-commit-config.yaml` to include ruff and exclude pylint.
- Made changes to files to handle changes which were detected by ruff.
- Removed pylint fully.
- **D1** needs to be discussed in detail if we want to ignore them currently or make changes by adding public docstrings.

Don't know why this test fails 🤔

## Checklist


preliz/distributions/distributions.py
```diff
@@ -204,11 +193,11 @@ def mode(self):
         raise NotImplementedError
 
     def std(self):
-        """Standard deviation."""
+        """Calculate Standard Deviation."""
```
aloctavodia, line 196: Standard deviation.

preliz/distributions/betabinomial.py
```diff
@@ -83,7 +79,7 @@ def pdf(self, x):
         return np.exp(self.logpdf(x))
 
     def cdf(self, x):
-        if isinstance(x, (np.ndarray, list, tuple)):
+        if isinstance(x, np.ndarray | list | tuple):
```
aloctavodia, line 82: Is this change requiered by ruff?

preliz/distributions/distributions.py
aloctavodia: We don't use calculate for the other properties/moments

preliz/distributions/betabinomial.py
rohanbabbar04: Yup, this was suggested by ruff UP038

preliz/distributions/distributions.py
rohanbabbar04: This is done, added `noqa: D401` as I was getting a ruff error here

preliz/distributions/betabinomial.py
rohanbabbar04: Do you think we should go ahead with correcting the UP038, or just ignoring it?

aloctavodia: Let's keep the union type.

aloctavodia (approved):

Issue #604, opened 2024-12-12
Add mode method


aloctavodia:
In #603  we introduced a mode argument to maxent and a mode method to many distributions. The mode is still missing from the following distributions.

* [x] AsymmetricLaplace
* [x] BetaScaled
* [x] ChiSquared
* [x] ExGaussian
* [x] Kumaraswamy
* [x] LogLogistic
* [x] LogitNormal
* [x] Pareto
* [x] Rice
* [x] SkewNormal
* [x] SkewStudentT
* [x] StudentT
* [x] TruncatedNormal
* [x] Uniform
* [x] VonMises
* [x] Wald
* [x] Weibull
* [x] BetaBinomial
* [x] Binomial
* [x] Categorical
* [x] DiscreteUniform
* [x] DiscreteWeibull
* [x] HyperGeometric
* [x] ZeroInflatedBinomial
* [x] ZeroInflatedNegativeBinomial
* [x] ZeroInflatedPoisson
* [x] Dirichlet
* [x] MultiVariateNormal
* [ ] Hurdle
* [ ] Censored
* [x] Truncated
* [ ] Mixture

rohanbabbar04:
I have a question: there are some distributions that are bimodal. Should we return both modes or just one?

aloctavodia:
Good question. Maybe only the major mode? Not sure what to do if two equal modes. 

Also for the beta and beta scaled, I think we are currently returning 0.5 when alpha and beta are lower than 1. But we should do:

0.5  if α=β = 1
No mode if α<1 OR β<1.
at 0 for α ≤ 1, and at 1 if β ≤ 1

aloctavodia:
yep, probably the better thing is to respect multimodality. Then we will need to check that setting the mode works for `maxent` or that at least a warning is raised if we can not handle those cases.

Advaitgaur004:
> One option is to go with the major one, other is to return a tuple of both modes

I believe returning a list or tuple of modes is beneficial. For a bimodal normal mixture, both peaks should be returned. That’s my suggestion.

Advaitgaur004:
> yep, probably the better thing is to respect multimodality. Then we will need to check that setting the mode works for `maxent` or that at least a warning is raised if we can not handle those cases.

I have now implemented unimodal modes, but for bimodal, we need to return tuples.

PR #632, opened 2025-01-21
Update Beta and BetaScaled and Add Weibull

rohanbabbar04:
#604 

## Description
- Add conditions to Beta and BetaScaled

- [x] Mode to Weibull Distribution

## Checklist

### DIFF
--- a/preliz/distributions/beta.py
+++ b/preliz/distributions/beta.py
@@ -249,7 +249,7 @@ def nb_neg_logpdf(x, alpha, beta):
     return -(nb_logpdf(x, alpha, beta)).sum()
 
 
-@nb.njit(cache=True)
+@nb.vectorize(nopython=True, cache=True)
 def nb_mode(alpha, beta):
     if alpha == 1 and beta == 1:
         return 0.5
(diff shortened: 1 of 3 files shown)


preliz/distributions/beta.py
```diff
@@ -165,9 +165,15 @@ def mean(self):
         return self.alpha / (self.alpha + self.beta)
 
     def mode(self):
-        return np.where(
-            (self.alpha > 1) & (self.beta > 1), (self.alpha - 1) / (self.alpha + self.beta - 2), 0.5
-        )
+        if self.alpha == 1 and self.beta == 1:
```
aloctavodia, line 168: We support parameters to be vectors, this will fail in those cases. You should explicitly write a vectorized form or use numba to vectorize a version working with scalars

rohanbabbar04: Oops...Completely forgot about that.
Fixed it :)

preliz/distributions/beta.py
```diff
@@ -249,3 +247,16 @@ def nb_logpdf(x, alpha, beta):
 @nb.njit(cache=True)
 def nb_neg_logpdf(x, alpha, beta):
     return -(nb_logpdf(x, alpha, beta)).sum()
+
+
+@nb.njit(cache=True)
```
aloctavodia, line 252:
```suggestion
@nb.vectorize(nopython=True, cache=True)
```

PR #634, opened 2025-01-23
Distributions Gallery: Separate Colors by Category

aleicazatti:
## Description

Updates the script for generating cover images to assign colors by category (continuous, discrete, modifier) instead of randomly. It’s a simple solution for now to make the colors meaningful.

This idea was mentioned in [#479](https://github.com/arviz-devs/preliz/pull/479), so it builds on that discussion.


docs/get_cover_gallery.py
```diff
@@ -7,11 +7,70 @@
 style.use("preliz-doc")
 rng = np.random.default_rng(247)
 
+category_colors = {
```
aloctavodia, line 10: Any reason for this particular order of the colors?

docs/get_cover_gallery.py
```diff
@@ -7,11 +7,70 @@
...
+    'continuous': 'C3',
+    'discrete': 'C1',
+    'modifier': 'C2'
+}
+
+distribution_categories = {
```
aloctavodia, line 16: No need to do this manually. All distributions are subclasses of classes matching this color mapping.

aloctavodia (changes requested):

docs/get_cover_gallery.py
aleicazatti: No reason. I just liked that order. But we can change it

docs/get_cover_gallery.py
aleicazatti: Working on it

PR #635, opened 2025-01-23
Add mode to Binomial

rohanbabbar04:
## Description

- [x] Binomial

## Checklist

### DIFF
--- a/preliz/distributions/binomial.py
+++ b/preliz/distributions/binomial.py
@@ -108,9 +108,9 @@ class Binomial(Discrete):
 
     def mode(self):
         y = (self.n + 1) * self.p
-        if (np.isin(self.p, [0, 1])) | (np.mod(y, 1) != 0):
-            return np.where(self.p == 1, self.n, np.floor(y))
-        return y, y - 1
+        if np.isscalar(y) and y.is_integer() and 0 < self.p < 1:
+            return y, y - 1
+        return np.where(self.p == 1, self.n, np.floor(y))
 
     def median(self):
         return np.ceil(self.n * self.p)
--- a/preliz/distributions/binomial.py
+++ b/preliz/distributions/binomial.py
@@ -108,8 +108,8 @@ class Binomial(Discrete):
 
     def mode(self):
         y = (self.n + 1) * self.p
-        if np.issubdtype(type(y), np.integer):
-            return (self.n + 1) * self.p, (self.n + 1) * self.p - 1
+        if not np.mod(y, 1):
+            return y, y - 1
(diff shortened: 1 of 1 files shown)


rohanbabbar04:
One thing to ask here:
In some cases, the binomial distribution is bimodal. Should we start handling the bimodalirity in this PR? Also, if we have any corrections for `maxent`, let's include them in this PR itself...

Advaitgaur004:
I am adding mode in other python files @rohanbabbar04 will make a pr.

rohanbabbar04:
> I am adding a mode to other Python files, and I will make a pull request. is that fine @rohanbabbar04

Sure, can you comment on #604 so that we don't pick the same ones?

Advaitgaur004:
> > I am adding a mode to other Python files, and I will make a pull request. is that fine @rohanbabbar04
> 
> Sure, can you comment on #604 which ones are you working on so that we don't pick the same ones?

Sure, you have already implemented Rice and Binomial, so I will drop this since I have already implemented it in my local setup.

aloctavodia:
> Some things to ask here:
> 
> 1. In some cases, the binomial distribution is bimodal. Should we focus on the major mode only?
> 
> image
> 2. For bimodal or multimodal like we discussed is it fine to return a tuple of modes and handle it in optimize max ent.
> 
> What do you suggest @aloctavodia ?

Not sure, but probably better to return a tuple if more than one mode. And as a first approach allow to fix a single mode in maxent, but handle de case that the mode method could return a tuple

rohanbabbar04:
Numba vectorize doesn't support `Tuple` to be returned, is there a possible way to include this

```python
def mode(self):
    y = (self.n + 1) * self.p
    if (np.isin(self.p, [0, 1])) | (np.mod(y, 1) != 0):
        return np.where(self.p == 1, self.n, np.floor(y))
    return y, y - 1
```

aloctavodia:
Not sure what to do. Having to deal with multiple modes is quite annoying and I am not sure it is worth the effort.

rohanbabbar04:
Hmm...
Should we go with the major mode instead. We show only one mode which is most likely to be major?

aloctavodia:
How weird would be two return two modes (if they exists) only if all parameters are scalar. And return only one for vector parameters. We support vector parameters, but we mostly care of scalars

rohanbabbar04:
I think the latest commit should work for both scalar and vectors(with scalar giving bimodal result)
What do you think @aloctavodia ?

aloctavodia (approved):

PR #636, opened 2025-01-23
Add mode method for DiscreteWeibull, LogitNormal, Hypergeometric and SkewNormal

Advaitgaur004:
## Description
This PR targeted #604 and work is in progress.

Implemented Modes 
- [x] Discreteweibull
- [x] logitnormal
- [x] hypergeometric
- [x] SkewNormal


Advaitgaur004:
- [ ] Binomial Distribution:
Becomes bimodal when n is large and p is 0.5
- [ ] Beta Distribution:
Becomes bimodal when both shape parameters (α and β) are less than 1
- [ ] Mixture Distribution:
Explicitly created by combining two Gaussian distributions
- [ ] Beta-Binomial Distribution:
Bimodality occurs with small α and β parameters

@rohanbabbar04 These need to be changed if we are opting to send a tuple for multimodal purposes.

preliz/distributions/discrete_weibull.py
```diff
@@ -107,6 +107,9 @@ def skewness(self):
     def kurtosis(self):
         return num_kurtosis(self)
 
+    def mode(self):
+        return np.floor(self.q * np.power((self.beta - 1) / self.beta, 1 / self.beta))
+
```
aloctavodia, line 112: This does not seem to be the correct expression. Where did you find it?

preliz/distributions/exgaussian.py
```diff
@@ -123,6 +123,9 @@ def kurtosis(self):
         opnus2 = 1.0 + nus2
         return 6.0 * nus2 * nus2 * opnus2 ** (-2)
 
+    def mode(self):
+        return self.mu
```
aloctavodia, line 127: This is also wrong. It will only be true if nu goes to 0, then you get a normal

preliz/distributions/logitnormal.py
```diff
@@ -149,6 +149,9 @@ def kurtosis(self):
         pdf = self.pdf(x_values)
         return np.trapz(((x_values - mean) / std) ** 4 * pdf, x_values) - 3
 
+    def mode(self):
+        return 1 / (1 + np.exp(-self.mu))
+
```
aloctavodia, line 154: This can't be true, the mode of the logitnormal may be affected by sigma

preliz/distributions/skewnormal.py
```diff
@@ -150,6 +150,10 @@ def kurtosis(self):
             * ((delta * np.sqrt(2 / np.pi)) ** 4 / (1 - 2 * (delta**2) / np.pi) ** 2)
         )
 
+    def mode(self):
```
aloctavodia, line 153: This looks like the expression for the mean, not mode

preliz/distributions/exgaussian.py
Advaitgaur004: I am bit a confused as mode of ExGaussian is 

image
and eliminating [complementary error function (erfc) ](https://en.wikipedia.org/wiki/Complementary_error_function) so we are end up with 

`mu + sigma^2 * nu` would be the mode, Is that acceptable, or do further changes need to be made?

preliz/distributions/discrete_weibull.py
Advaitgaur004: Yes, this is a bit off.
image
mode for Weibull is this, where `k = alpha` and `lambda = beta` .

Advaitgaur004: On approval i can change the function correctly

preliz/distributions/logitnormal.py
Advaitgaur004: it is bimodal if i am not wrong. so we have to return tuples of values, and modes should be,

preliz/distributions/exgaussian.py
aloctavodia: If you are asking if we can compute the mode as `mu + sigma**2 * nu`, then no. That's not the mode. If you are unsure about how to compute the mode, better to drop it. If you want to compute the mode, then find a reliable source and implement it.

preliz/distributions/skewnormal.py
Advaitgaur004: Yes, the mode should be mu + sigma*mo(alpha)
where mo(alpha) is 
image

it is very complex to be honest, how to approach this.

preliz/distributions/skew_studentt.py
```diff
@@ -220,6 +220,15 @@ def kurtosis(self):
...
 
+    def mode(self):
+        # For α=0 (symmetric case), mode = ξ
+        # For other cases, mode needs numerical optimization
+        return np.where(
+            self.a == 0,
+            self.mu,
+            self.mu  # This is an approximation,
+        )
+
```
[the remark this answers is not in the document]
Advaitgaur004: okay i'll remove this for the moment

preliz/distributions/logitnormal.py
Advaitgaur004: one way is to assume that x is in between 0 and 1 and assuming logit(x) = log(x/1-x)
```python
def logit(x):
    if x == 0:
        return -np.inf
    elif x == 1:
        return np.inf
    if x < 0 or x > 1:
        return np.nan
    else:
        return np.log(x / (1 - x))
```

still end up at 
image
and this required approximations

aloctavodia: again, find a reliable source and implement that.

preliz/distributions/skewnormal.py
aloctavodia: It is what it is

preliz/distributions/exgaussian.py
Advaitgaur004: I understand how to calculate this, but it becomes complicated. Ok I'll find a reliable source then

preliz/distributions/logitnormal.py
Advaitgaur004: Ok cap

Advaitgaur004:
@aloctavodia Issues have been updated for discrete_weibull; I will use minimize_scale from scipy.optimize since direct calculation for mode in discrete_weibull isn't feasible. For the rest, the mode is calculated as provided in Wikipedia or some reliable sources.

preliz/distributions/discrete_weibull.py
```diff
@@ -107,6 +107,14 @@ def skewness(self):
     def kurtosis(self):
         return num_kurtosis(self)
 
+    def mode(self):
```
aloctavodia, line 110: optimization routines should be inside optimization module, as they can be potentially reused. The 100 is a magic number, maybe a better option is to use self.ppf(0.9999) or similar. I tried with a couple of examples and the mode is not always right,

preliz/distributions/exgaussian.py
```diff
@@ -1,6 +1,7 @@
 import numba as nb
 import numpy as np
 from scipy.stats import skew
+from scipy.special import erfcinv
```
aloctavodia, line 4: There is an `erfcinv` function implemented in the special submodule

preliz/distributions/exgaussian.py
```diff
@@ -101,6 +102,14 @@ def entropy(self):
         logpdf = self.logpdf(x_values)
         return -np.trapz(np.exp(logpdf) * logpdf, x_values)
 
+    def mode(self):
+        tau = self.nu
```
aloctavodia, line 106: why renamming  the variable? I get nan for some combinations of parameters like pz.ExGaussian(0, 1, 10)

preliz/distributions/logitnormal.py
```diff
@@ -149,6 +149,37 @@ def kurtosis(self):
         pdf = self.pdf(x_values)
         return np.trapz(((x_values - mean) / std) ** 4 * pdf, x_values) - 3
 
+    def mode(self):
```
aloctavodia, line 152: This should be in the optimization module. Do you have a reference for this method? It seems to work fine for the values I manually checked. Maybe you can check how the mode is numerically computed here https://github.com/bgctw/logitnorm

preliz/distributions/skew_studentt.py
```diff
@@ -220,6 +220,10 @@ def kurtosis(self):
         )
         return nu4 / nu2**2 - 3
 
+    def mode(self):
+        #Not implement yet
+        pass
+
```
aloctavodia, line 226:
```suggestion
```

preliz/distributions/skewnormal.py
```diff
@@ -150,6 +150,21 @@ def kurtosis(self):
...
```
aloctavodia, line 153: This gives the wrong results for large absolute values of alpha. Not sure about the proper fix, but maybe it could be  enough to ensure the mode is no smaller than mu (for positive values of alpha) and no larger than mu (for negative values of alpha)

aloctavodia (changes requested):

aloctavodia:
@Advaitgaur004 are you still working on this?

Advaitgaur004:
> @Advaitgaur004 are you still working on this?

Yes, i am busy for 2 more days, i'll made pr in some time

aloctavodia:
great! Thanks

Advaitgaur004:
Changes Made in 
```
Discreteweibull
Exgaussian
logitnormal
hypergeometric
SkewNormal
```

preliz/internal/optimization.py
```diff
@@ -452,6 +452,32 @@ def solve_for_kappa(kappa):
         return np.finfo(float).tiny
 
 
+def find_mode_through_optimization(dist, bounds=None):
```
aloctavodia, line 455:
```suggestion
def find_mode(dist, bounds=None):
```

preliz/internal/optimization.py
```diff
@@ -452,6 +452,32 @@ def solve_for_kappa(kappa):
...
+    Returns
+    -------
+    float
+        Mode of the distribution
+    """
+    def negative_pdf(x):
+        return -dist.pdf(x)
+    
+    if bounds is None:
+        bounds = (0, dist.ppf(0.9999))
```
aloctavodia, line 475:
```suggestion
        bounds = dist._finite_endpoints("full")
```

preliz/distributions/exgaussian.py
```diff
@@ -101,6 +101,10 @@ def entropy(self):
         logpdf = self.logpdf(x_values)
         return -np.trapz(np.exp(logpdf) * logpdf, x_values)
 
+    def mode(self):        
+        t = min((np.abs(self.nu)/self.sigma) * np.sqrt(2/np.pi), 1.99)
+        return self.mu - np.sign(self.nu) * np.sqrt(2*self.sigma) * erfcinv(t) + self.sigma**2/self.nu
+
```
aloctavodia, line 107: This is not right, please do some sanity checks, like 

```{python}
dist = pz.ExGaussian(0, 10, 2)
ax = dist.plot_pdf()
ax.axvline(dist.mode(), ls='--', color='k')
```

you can drop this one if you want

preliz/distributions/logitnormal.py
aloctavodia, line 152: Move this to the optimization module and rename to something line find_mode_logitnormal

aloctavodia (changes requested):

preliz/distributions/exgaussian.py
Advaitgaur004: I found the mode from Wikipedia. I'm dropping this due to the lack of an authentic source.

aloctavodia: ok

preliz/distributions/logitnormal.py
```diff
@@ -1,6 +1,6 @@
 import numba as nb
 import numpy as np
-
+from scipy.optimize import root_scalar
```
aloctavodia, line 3:
```suggestion
```

preliz/distributions/skew_studentt.py
```diff
@@ -265,4 +265,4 @@ def nb_logpdf(x, mu, sigma, a, b):
 
 @nb.njit(cache=True)
 def nb_neg_logpdf(x, mu, sigma, a, b):
-    return -(nb_logpdf(x, mu, sigma, a, b)).sum()
+    return -(nb_logpdf(x, mu, sigma, a, b)).sum()
```
aloctavodia, line 268: This is unrelated to this PR

preliz/distributions/continuous_multivariate.py
```diff
@@ -668,4 +668,4 @@ def plot(**args):
                     xy_lim,
                 )
 
-        return interactive(plot, **plot_widgets)
+        return interactive(plot, **plot_widgets)
```
aloctavodia, line 671: Unrelated to this PR

aloctavodia (changes requested):

preliz/distributions/continuous_multivariate.py
Advaitgaur004: same with this

preliz/distributions/studentt.py
```diff
@@ -240,4 +240,4 @@ def nb_logpdf(x, nu, mu, sigma):
 
 @nb.njit(cache=True)
 def nb_neg_logpdf(x, nu, mu, sigma):
-    return -(nb_logpdf(x, nu, mu, sigma)).sum()
+    return -(nb_logpdf(x, nu, mu, sigma)).sum()
```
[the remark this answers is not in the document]
Advaitgaur004: same again

aloctavodia:
@Advaitgaur004 check the tests that are failing, you will see linter errors.

Advaitgaur004:
> @Advaitgaur004 check the tests that are failing, you will see linter errors.

I've finished reviewing the files. Now I see the issue with those three files; I didn't apply the black style, which caused the problem. I have now updated all the files accordingly, and there are no linter errors.

PR #637, opened 2025-01-27
Update dependencies

rohanbabbar04:
## Description
- Remove fixed dependencies from requirements.(Would be better for future versions of python).

## Checklist

- [x] All Tests pass


rohanbabbar04:
- I think we should get rid of the fixed dependencies `(==)` which will make it easier to migrate to python3.13.
- Don't really know why this error is coming(could be due to some minor adjustments in random values)

rohanbabbar04:
> We can get small differences like this due to changes from one scipy version to the other. Dependencies can be tricky. If we leave them unfixed, users may get errors or missing features on new installations just because a dependency got updated, but untested on our side. We should probably stick to this https://scientific-python.org/specs/spec-0000/

Yup, I have updated to `scipy>=1.10, <1.15` which seems to be good, our tests work fine as well

rohanbabbar04:
@aloctavodia, this is ready to be reviewed.
With this it should be possible to incorporate python3.13 without much fuss :)

PR #650, opened 2025-02-21
Add mode function to Continuous class

rohanbabbar04:
## Description
- Should cover up remaining distributions in Continuous which don't have an analytical expression

## Checklist


aloctavodia:
We can merge, but I would like to manually check this is producing sensible results. Can you list the distributions I should check?

rohanbabbar04:
These are the 4 distributions:
- ExGaussian
- Rice
- SkewStudentT
- SkewNormal

aloctavodia:
Thanks!

PR #651, opened 2025-02-22
Add sphinxbibtex and function to get references from methods

### LINKED ISSUE #640: How to cite function and page
We could list on a single page not only how to cite PreliZ but also the methods implemented in PreliZ. Additionally, we could have a function that prints the citations maybe in BibTeX format to the screen and/or save them to a file. By default, the function could return just Preliz's citation, but we could pass an argument to list everything or pass a list a method to return just the requested citations.

rohanbabbar04:
Closes #640 
- Add sphinxcontrib.bibtex to handle references.
- Add function to generate citations.

## Checklist


rohanbabbar04:
This DOI in [mle](https://preliz.readthedocs.io/en/latest/unidimensional.html#preliz.unidimensional.mle) looks broken.

aloctavodia:
This is the correct one https://doi.org/10.1177/0049124104268644

preliz/utils/citations.py
```diff
@@ -0,0 +1,51 @@
+import os
+import re
+
+try:
+    from IPython.display import Markdown, display
+except ImportError:
+    pass
+
+
+def citations(methods=None, show_as="bibtex"):
```
aloctavodia, line 10: This should also work outside a notebook environment

preliz/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from preliz.utils.citations import citations
```
aloctavodia, line 1: "internal"  is our "utils"

preliz/utils/citations.py
rohanbabbar04: In that case, will it return a string of the citations?

aloctavodia: yes, we should also have an option to write to a file

aloctavodia (approved):

PR #660, opened 2025-03-26
Add mode method for Discrete distributions

aleicazatti:
## Description

- This PR adds `find_discrete_mode` in `internal.optimization`. It can be used in the remaining discrete distributions which don't have an analytical or simple solution for the mode. It evaluates the PMF over the distribution's support and returns the value with the highest probability.
- Applied this helper to the mode() method of ZeroInflatedBinomial, ZeroInflatedNegativeBinomial and ZeroInflatedPoisson.

## Checklist

- [X] Code style is correct (follows ruff and black guidelines)

### DIFF
--- a/preliz/distributions/zi_binomial.py
+++ b/preliz/distributions/zi_binomial.py
@@ -6,6 +6,7 @@ from preliz.distributions.distributions import Discrete
 from preliz.internal.distribution_helper import all_not_none, eps
 from preliz.internal.optimization import optimize_ml, optimize_moments
 from preliz.internal.special import cdf_bounds, gammaln, ppf_bounds_disc
+from preliz.distributions.binomial import Binomial
 
 
 class ZeroInflatedBinomial(Discrete):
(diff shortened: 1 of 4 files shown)


preliz/distributions/zi_binomial.py
```diff
@@ -101,6 +102,28 @@ def entropy(self):
...
+        else:
+            prob_zero = (1 - self.psi) + self.psi * (1 - self.p) ** self.n
+            prob_binomial_mode = self.psi * np.exp(
+                gammaln(self.n + 1)
+                - gammaln(binomial_mode + 1)
+                - gammaln(self.n - binomial_mode + 1)
+                + binomial_mode * np.log(self.p)
+                + (self.n - binomial_mode) * np.log1p(-self.p)
+            )
+            return 0 if prob_zero > prob_binomial_mode else binomial_mode
```
aloctavodia, line 125: I would write this as 

```suggestion
        if self.psi == 0:
            return 0
        else:
            bin_dist = Binomial(self.n, self.p) 
            binomial_raw_mode = bin_dist.mode()
            binomial_mode = (
                min(binomial_raw_mode)
                if isinstance(binomial_raw_mode, tuple)
                else int(binomial_raw_mode)
            )

            if self.psi == 1:
                return binomial_mode
            else:
                prob_zero = (1 - self.psi) + self.psi * (1 - self.p) ** self.n
                prob_binomial_mode = self.psi * bin_dist.pdf(binomial_mode)
                return 0 if prob_zero > prob_binomial_mode else binomial_mode
```

But then I realized I was confused about how we have implemented ZIB. When we talked, I had the wrong idea that we have explicitly impleted as a binomial + a zero_inflation.  So may we should use the pmf method you suggested like this:

        x_vals = self.xvals("full")
        pmf_vals = self.pdf(x_vals)
        return x_vals[np.argmax(pmf_vals)].astype(int)
        
This should be implemented as a general method for all the missing discrete variables (unless we have a more direct way for them).

aloctavodia:
Could you add this to all the discrete distributions without a mode method? We can later replace some of them with a specific method if available or easy-to-implemet

PR #663, opened 2025-04-24
Add Mode method to Truncated

aleicazatti:
## Description

- Adds mode method to the `Truncated` distribution modifier, using `find_mode` or `find_discrete_mode` depending on the base distribution type.

## Checklist

- [X] Code style is correct (follows ruff and black guidelines)

### DIFF
--- a/preliz/internal/optimization.py
+++ b/preliz/internal/optimization.py
@@ -515,7 +515,7 @@ def find_mode(distribution, bounds=None):
     return result.x
 
 
-def find_discrete_mode(dist):
+def find_discrete_mode(dist, bounds=None):
     """
     Find mode for a discrete distribution from its pmf.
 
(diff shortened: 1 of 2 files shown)


preliz/internal/optimization.py
```diff
@@ -515,21 +515,25 @@ def negative_pdf(x):
...
-def find_discrete_mode(dist):
+def find_discrete_mode(dist, bounds=None):
     """
     Find mode for a discrete distribution from its pmf.
 
     Parameters
     ----------
     dist : Distribution
         PreliZ distribution object
+    bounds : tuple, optional
```
aloctavodia, line 526: Are these bounds used anywhere? When do we want to use them? The same goes for `find_mode`. I am asking because `xvals` already takes into account the support of the distribution. The bounds could be used if we know the mode is in a region smaller than the support of the distribution, but I think we currently don't have any case like that.

Issue #683, opened 2025-08-28
Categorical distribution samples from probabilities instead of categories


fietensen:
**Describe the bug**
The `Categorical` distribution’s `rvs` method is incorrectly implemented.
Currently it samples from the raw probability values in `p` instead of the category indices `{0, …, len(p)-1}` with probabilities `p`. This leads to invalid samples (e.g. returning `0.6` or `0.3` instead of `0` or `2`).

**To Reproduce**
```python
import numpy as np
from preliz.distributions import Categorical

p = [0.1, 0.6, 0.3]
dist = Categorical(p)

samples = dist.rvs(size=10, random_state=42)
print(samples) # returns values from [0.1, 0.6, 0.3], not {0, 1, 2}
```

**Expected behavior**
`rvs` should return integer indices of categories, sampled according to `p`. For the example above, results should look like [1, 2, 1, …] with frequencies approximately matching [0.1, 0.6, 0.3].

A corrected implementation could be:
```py
    def rvs(self, size=None, random_state=None):
        random_state = np.random.default_rng(random_state)
        support_array = np.arange(self.support[0], self.support[1] + 1)
        return random_state.choice(support_array, size, p=self.p)
```
The support array enforcing the `choice` method to return an `int64` object for `size==None` instead of a `numpy` array as to be consistent with other discrete distributions such as `DiscreteUniform`. ((although it might actually be more consistent to wrap **all** discrete distribution's results in `np.atleast_1d` in order to make `rvs` consistently return `numpy` arrays))

**Additional context**
preliz version `0.20.0`

PR #684, opened 2025-08-28
Fix `Categorical.rvs` to sample categories using probabilities

fietensen:
## Description
This PR fixes the implementation of the `rvs` method in the Categorical distribution. Previously, `rvs` incorrectly sampled directly from the probability vector `p`, returning probability values instead of the category indices `{0, …, len(p)-1}`.

The corrected implementation uses `numpy.random.Generator.choice` to sample indices according to the weights `p`. 

This resolves issue #683 and ensures that random variates are consistent with the specified categorical distribution.

## Checklist


aloctavodia:
Thanks a lot for this fix!

PR #720, opened 2026-05-19
 Fix plot_interactive in MvNormal and Dirichlet 

rohanbabbar04:
## Description
Looks like there is an issue with `plot_interactive` method, because of mismatch in positional arguments.
- `baseline` is missed, leading to errors.

## Checklist


rohanbabbar04:
The docs are failing, I think it is related to some issue in `bambi`.

PR #725, opened 2026-06-09
Add catalog, richer html representation and info method

aloctavodia:


preliz/tests/t_dd8b0f8cb147459a970a4fb65d777238.py
rohanbabbar04, line 1: I think this was pushed accidentally, we can remove this 😅

preliz/distributions/catalog.py
```diff
@@ -0,0 +1,403 @@
...
+        >>> pz.dists.names()
+        >>> pz.dists.names("discrete")
+
+    Get info about a specific distribution::
+
+        >>> pz.dists.info("Gamma")
+
+    Filter distributions by properties::
+
+        >>> pz.dists.find(kind="continuous", num_params=2)
```
rohanbabbar04: All of them should be `pz.catalog.func`

rohanbabbar04 (commented):
Looks great!!
Left some minor comments, after which we can merge this.

preliz/distributions/catalog.py
rohanbabbar04: For `pz.dists` to work we can alias in `__init__.py`, so we don't need to change var name from catalog to dists.
```
from preliz.distributions.catalog import catalog as dists
```

I would also prefer that `pz.dists` is exposed to the users instead of `pz.catalog`

aloctavodia: I started this as a `get_dists` function, and then moved into the class, and changing the name to catalog was one of the last changes, hehe.

rohanbabbar04 (approved):

rohanbabbar04:
Did some very minor changes :sweat_smile: , this is a great addition!!!, should be ready to be merged.

aloctavodia:
Thanks @rohanbabbar04!

PR #727, opened 2026-06-18
Fix import failure on matplotlib 3.11

### LINKED ISSUE #726: `import preliz` fails on matplotlib 3.11 (`module 'matplotlib.style' has no attribute 'core'`)
**Describe the bug**

`import preliz` fails with matplotlib 3.11.0. The line `style.core.USER_LIBRARY_PATHS.append(...)` in `preliz/__init__.py` runs at import time and raises `AttributeError`. matplotlib 3.11 no longer auto-imports the `matplotlib.style.core` submodule, so the `style.core` attribute access fails (the submodule still exists, but `import matplotlib.style.core` is now required to reach it). This affects 0.24.0, 0.25.0, 0.26.0 (latest), and `main`.

**To Reproduce**

```
pip install "matplotlib==3.11.0" "preliz==0.26.0"
python -c "import preliz"
```

Stack trace:

```
File ".../preliz/__init__.py", line 31, in <module>
    style.core.USER_LIBRARY_PATHS.append(_preliz_style_path)
AttributeError: module 'matplotlib.style' has no attribute 'core'
```

**Expected behavior**

`import preliz` works on matplotlib 3.11. A fix mirroring the one arviz-plots used (use the new top-level location, fall back to `.core` for old matplotlib):

```python
if hasattr(style, "USER_LIBRARY_PATHS"):
    style.USER_LIBRARY_PATHS.append(_preliz_style_path)
else:
    style.core.USER_LIBRARY_PATHS.append(_preliz_style_path)
style.reload_library()
```

**Additional context**

matplotlib 3.11.0, preliz 0.26.0, Python 3.12, Linux. Found while trying to drop a `matplotlib<3.11` pin in pymc-marketing, where preliz is imported lazily via `pymc_extras.prior.Prior.constrain`.

I reproduced it with uv in a throwaway environment:

```
uv run --python 3.12 --with "matplotlib==3.11.0" --with "preliz==0.26.0" python -c "import preliz"
```

anevolbap:
## Description

matplotlib 3.11 no longer auto-imports the `matplotlib.style.core` submodule, so the `style.core.USER_LIBRARY_PATHS.append(...)` call in `preliz/__init__.py` raises `AttributeError` at import time. This uses the top-level `style.USER_LIBRARY_PATHS` when it exists, falls back to `style.core` for older matplotlib, and calls `style.reload_library()` (present on all supported versions).

Checked across matplotlib 3.9, 3.10, and 3.11: `style.USER_LIBRARY_PATHS` exists only on 3.11, `style.core` exists on 3.9 and 3.10, and `style.reload_library` exists on all three, so the guard covers the full `matplotlib>=3.9` range.

Closes #726.

## Checklist

### DIFF
--- a/preliz/__init__.py
+++ b/preliz/__init__.py
@@ -29,8 +29,11 @@ mpl_rcParams["savefig.bbox"] = "tight"
 
 # add PreliZ's styles to matplotlib's styles
 _preliz_style_path = os_path.join(os_path.dirname(__file__), "styles")
-style.core.USER_LIBRARY_PATHS.append(_preliz_style_path)
-style.core.reload_library()
+if hasattr(style, "USER_LIBRARY_PATHS"):
+    style.USER_LIBRARY_PATHS.append(_preliz_style_path)
+else:
+    style.core.USER_LIBRARY_PATHS.append(_preliz_style_path)
+style.reload_library()
 
 # clean namespace
 del os_path, mpl_rcParams, _preliz_style_path


aloctavodia:
thanks @anevolbap

PR #731, opened 2026-08-10
Add Wishart distribution

rohanbabbar04:
## Description

- Add Wishart distribution
- Update pytensor_jit and pytensor_rng_jit to handle `constant`.
- Add test for Wishart distribution
- Add `plot_wishart`, the plotting only works when V is 1x1(p=1)

## Checklist


preliz/distributions/distributions.py
```diff
@@ -52,16 +52,21 @@ def _get_description(self):
...
-            (
-                f"{n}={v:.3g}, "
-                if np.isscalar(v) or np.ndim(v) == 0
-                else f"{n}=["
-                + "".join(f"{vi:.3g}, " for vi in np.atleast_1d(v)).strip(", ")
-                + "], "
-            )
-            for n, v in zip(self.param_names, self.params)
-        ).strip(", ")
+        def fmt(v):
```
aloctavodia: why do we need this?

preliz/distributions/wishart.py
```diff
@@ -0,0 +1,157 @@
...
+        random_state = np.random.default_rng(random_state)
+        return ptd_rvs(self.nu, self.V, size=size, rng=random_state)
+
+    def _fit_moments(self, mean, sigma):
+        diag_mean = np.diagonal(mean, axis1=-2, axis2=-1) if mean.ndim > 1 else mean
+        diag_sigma = np.diagonal(mean, axis1=-2, axis2=-1) if sigma.ndim > 1 else sigma
+        nu_estimates = 2 * (diag_mean / diag_sigma) ** 2
+        nu = np.mean(nu_estimates)
+        V = mean / nu
+        print(V, nu)
```
aloctavodia:
```suggestion
```

preliz/internal/distribution_helper.py
```diff
@@ -105,93 +106,161 @@ def num_kurtosis(dist):
         return np.trapezoid(((x_values - mean) / std) ** 4 * pdf, x_values) - 3
 
 
-def pytensor_jit(func, **compile_kwargs):
+def pytensor_jit(func=None, *, constants=(), **compile_kwargs):
```
aloctavodia: Why do we need to change this function? How is it related to Wishart's implementation?

preliz/internal/plot_helper_multivariate.py
```diff
@@ -365,3 +367,75 @@ def plot_mvnormal(
...
+    support,
+    baseline,
+    legend,
+    figsize,
+    axes,
+    xy_lim="auto",
+):
+    """Plot pdf of a 1x1 (or batch of 1x1) Wishart distribution(s).
+
+    For p=1, the Wishart(V, nu) reduces exactly to a scaled chi-squared
```
aloctavodia, line 386: Do we want to have some kind of representation for p>1? Not sure there is one that is nice; even the marginals that we offer for MvNormal and Dirichlet, I think, are not as straightforward here.

preliz/internal/distribution_helper.py
```diff
@@ -105,93 +106,161 @@ def num_kurtosis(dist):
...
+        return inner_func
 
-    return inner_func
+    if func is not None:
+        return decorator(func)
+    return decorator
 
 
-def pytensor_rng_jit(_func=None, **compile_kwargs):
+def pytensor_rng_jit(_func=None, *, constants=(), **compile_kwargs):
```
aloctavodia: Same as the previous question.

aloctavodia (commented):
Do you have an idea of an elicitation method for the Dirichlet? Or some nice way to represent it (plotting for p>1)?

preliz/distributions/distributions.py
rohanbabbar04: Yup, no need for this, I was testing out something and pushed it. Have removed it.

preliz/internal/plot_helper_multivariate.py
rohanbabbar04: Yes, I was also looking for something which we can use to plot for p>1, didn't really come across anything straightforward, more or less every plot was when p was 1d.

preliz/internal/distribution_helper.py
rohanbabbar04: I have changed it slightly, I have added a `static_shapes` parameter which can be used to fixed the shape of the tensor. 
For Wishart, we would need to know the last axis of V (which is p) for it to compute.

